PR #15, opened 2015-01-20
added setter for custom UserAgent String.

JacobZ:


dotsbb:
@JacobZ if it is possible, please add test case.

JacobZ:
@dotsbb I added the test test.

While executing the Tests i had a problem with validateDefaultQuery. It assumes that die Language in the params is always en. In my case it was de (since I have a german setup on my machine). I just changed it for my tests but did not commit it. Maybe you can have a look on it sometime.

dotsbb:
@JacobZ Thanks, I'll try to mock `Locale.getDefault().getLanguage();` to make it working equally on every environment.

PR #19, opened 2015-03-16
Updated gradle & plugins so piwik-sdk builds on Android Studio 1.1

d4rken:
I updated gradle and all plugins so that you can import and build it in Android Studio 1.1 which is the current stable release.
The test command now requires **:piwik_sdk:build** but otherwise it works as previously. Additionally the unit tests can now also be run from within Android Studio.
> $ ./gradlew :piwik_sdk:clean **:piwik_sdk:build** jacocoTestReport generateReleaseJavadoc coveralls --info

The JavaDoc routine was changed to work with the newer gradle version.
And runProguard was removed from the demo app because it is no longer available (was set to false anyways)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,11 @@ jdk: oraclejdk7
 
 android:
   components:
-    - build-tools-19.1.0
-    - android-19
+    - build-tools-22.0.0
+    - android-22
     - extra-android-support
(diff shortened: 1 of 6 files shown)


d4rken:
I think this might need a few changes to the travis build script.

PR #20, opened 2015-03-16
Updates so the demo app builds against API22 and fixes for a few LINT errors.

d4rken:
* Moved config declarations out of manifest file into gradle file
* Set API & Tools to 22
* Removed proguard settings which were not used and caused issues on newer gradle versions
* Fixed LINT errors of type ERROR in the settings layout file (Sorry for the autoformat)

### DIFF
--- a/demo_app/build.gradle
+++ b/demo_app/build.gradle
@@ -1,20 +1,18 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 19
-    buildToolsVersion "19.1.0"
+    compileSdkVersion 22
+    buildToolsVersion "22"
 
     defaultConfig {
         applicationId "com.piwik.demo"
         minSdkVersion 11
-        targetSdkVersion 19
+        targetSdkVersion 22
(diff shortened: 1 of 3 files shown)


dotsbb:
Good job :+1: Also I like your [changes](https://github.com/d4rken/piwik-sdk-android/commit/109680144a8522feca7127eadc08141a328195fe) of QueryParams class.

d4rken:
I can send an extra pull request for the QueryParams changes.

Will you merge the other PR ( #19 ) so the lib builds with AS 1.1?

I would like to do some more fundamental changes besides  #23 #24 #25
Such as to remove the whole Piwik/Tracker differentation, because i don't see the usecase where we need different Tracker objects, except to deal with multiple threads and we could just make the Tracker threadsafe.
I would like to make Tracker an abstract class which you can extend and implement something like getTrackerUrl() to set the URL, then just do MyTracker.getInstance(...).doStuff.

The question is if this would take the library in a direction you don't want to take it and I should make my own fork.

PR #22, opened 2015-03-16
Store the visitor ID regex pattern as class variable

d4rken:
Such that it only has to be compiled once, little performance gain, but still.
Also changed Matcher.find() to Matcher.matches(), we don't need extra functionality of what "find()" does.


dotsbb:
Thanks for your contribution. Merged.

PR #29, opened 2015-03-24
Corrected resolution for API14+

d4rken:
The previous getResolution() implementation did not always yield correct results, the new solution gives the correct values for API14+. 

AFAIK there is no 100% solution for <API14 devices just via Context accessor without significant overhead.

Noticed it while testing on my Nexus5@5.1, reported was "1080x1776" actual resolution is "1080x1920".
The correct approach for this is only available since API17, for API14-16 we can trick a bit with reflection and if that fails (i.e.<API14), then we use the previous method.


dotsbb:
Thanks for your contribution.

Issue #24, opened 2015-03-17
Chaos in autoBindActivities


d4rken:
Currently onActivityStarted & onActivityResumed both end up calling activityStart(...), leading to 2 tracked screen views each time the activity starts (app starts / app rotation).

When onActivityPaused & onActivityStopped is called, they both call through to activityStop(...). In the same case as above we trying to dispatch two times directly after each other.
The documentation also states that stopAutoDispatching() would be called when the Activity is paused, but this is currently not happening.

I propose to change it like this:
onActivityStarted ->
onActivityResumed ->  start autodispatching & track screen view
onActivityPaused -> stop auto dispatching
onActivityStopped -> dispatch remaining events
onActivityDestroyed ->

dotsbb:
:+1: I'd definitely merge your pull request.

d4rken:
This is resolved, any further discussion can be done in #30.

PR #30, opened 2015-03-25
Autobinder refactoring

d4rken:
The autobind method now resides in a new class called "QuickTrack".
I've changed the autobinder calledbacks to force a dispatch on activity stop and otherwise just track a screenview "onResume".
It also returns the callback object (after registering it) so the dev can store the object and potentially unregister the callback again.
I've also moved it's helper functions into this class as they were not used for anything else.
The QuickTrack class could later be expanded with other things to "quickly track" ;).

I did not yet implement dispatch pausing as I'm currently uncertain wether it's the best way to go.

With this PR #24 is now resolved.


dotsbb:
I like your style, well documented and easy readable code!

Issue #28, opened 2015-03-23
Why track Exception with Piwik?


d4rken:
I'm looking for a use case or legitimization why we should track and forward exceptions.
Is there a plugin for the service tracker that visualizes them well and allows for organization?
Or to prevent reports from older apps for solved issue?

In bigger apps with 100k+ users, you always have crashs that are out of your control and after a few weeks the "Crash Event" list would be filled with useless stuff and no way to hide or remove these except manual database alternation. I do not see this being useful except maybe to have screenview tracking like start->screen1->screen2->crash. Although this is cool, with bigger usernumbers, i can't see this being favorable over traditional crash tracking (ACRA, Crashlytics etc.).

At least least this should be opt-in, not on by default.
But maybe this can be removed and we don't have to maintain it.

mattab:
> At least least this should be opt-in, not on by default.

:+1: 

>  But maybe this can be removed and we don't have to maintain it.

I think it could be useful to users to automatically track exception as [Events](http://piwik.org/docs/event-tracking/) in their app

d4rken:
Explicitly tracking exceptions could have some usecase, but even then it's almost always replaceable with if-checks, then track a normal event, which is better practice.

Automaticially tracking all exceptions needs strong filtering and extra options on the server side.
As i said for small apps 5k-20k active users this could work, but even then i would recommend picking something like Crashlytics that was made for this. Having options to hide processed exceptions, stopping reporting from older app versions etc.

If we venture into higher user counts like 100k-150k users, you will get daily crash that are unrelated to your app, but just got caught in your exception handler. With no server-side organisiation, it will spam the event section until it's unusuable.

I would propose to reduce the crash tracking to method that wraps the normal trackEvent method and just converts an Exception object to a useful message. Then remove everything else including automatic exception catching.

I don't want to throw the current piwik-android-sdk over, I'm just speaking out of experience. Sorry for all the scrutiny.

mattab:
Oh it's very useful to have your scrutiny actually!

> I would propose to reduce the crash tracking to method that wraps the normal trackEvent method and just converts an Exception object to a useful message. Then remove everything else including automatic exception catching.

Would this change the way the exception are tracked, eg. https://github.com/piwik/piwik-sdk-android#dispatching ? if so please submit the change to documentation with your PR

d4rken:
No, it will stay the same, I would basically remove the automatic exception handling and then simplify the exposed API.

[...]

PR #33, opened 2015-03-25
Exception handling refactoring

d4rken:
I've removed the uncaught exception handler due to concerns described in #28.
I've also removed the trackException method that took arbitrary strings as arguments to discourage overusing this or abusing it. In most cases we really want to track a normal event, exceptions are not replacement for if-clauses.

We are left with a trackException method that requires a throwable object. "Custom Exception" are still very much possible just create your own Exception or use a default Exception that takes arguments in the constructor.

I also updated the documentation for the method to point out that Piwik is not a crash tracker. Additionally i mentioned possibly pitfalls from proguard interferring with event matching due to obfuscated classnames or missing linenumbers.

I also contemplated whether the trackException method is really just another helper and should be moved into the new "QuickTrack" class, it's basically a wrapper, exception it just a name for an event.
This would be in contrast to i.e. downloads which are separately tracked on Piwiks UI.

Tracker: Methods for things that are distinct within Piwik UI.
QuickTrack: Methods for things that translate stuff into generic events in the Piwik UI.

This PR doesn't move it though yet, what do you think @dotsbb ?

### DIFF
--- a/demo_app/src/main/java/com/piwik/demo/DemoActivity.java
+++ b/demo_app/src/main/java/com/piwik/demo/DemoActivity.java
@@ -94,8 +94,7 @@ public class DemoActivity extends ActionBarActivity {
         ((PiwikApplication) getApplication()).getTracker()
                 .setDispatchInterval(5)
                 .trackAppDownload()
-                .setUserId(getUserId())
-                .reportUncaughtExceptions(true);
+                .setUserId(getUserId());
 
         initTrackViewListeners();
     }
@@ -127,7 +126,7 @@ public class DemoActivity extends ActionBarActivity {
         button.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
-                int a = 1 / 0;
+                ((PiwikApplication) getApplication()).getTracker().trackException(new Exception("OnPurposeException"), "Crash button", false);
             }
         });
 
@@ -142,7 +141,7 @@ public class DemoActivity extends ActionBarActivity {
                             ((EditText) findViewById(R.id.goalTextEditView)).getText().toString()
                     );
                 } catch (Exception e) {
-                    ((PiwikApplication) getApplication()).getTracker().trackException("DemoActivity", "wrong revenue", false);
+                    ((PiwikApplication) getApplication()).getTracker().trackException(e, "wrong revenue", false);
                     revenue = 0;
                 }
                 ((PiwikApplication) getApplication()).getTracker().trackGoal(1, revenue);
--- a/piwik_sdk/src/main/java/org/piwik/sdk/Piwik.java
+++ b/piwik_sdk/src/main/java/org/piwik/sdk/Piwik.java
@@ -32,8 +32,6 @@ public class Piwik {
 
     private boolean dryRun = false;
 
-    protected final static Thread.UncaughtExceptionHandler defaultUEH = Thread.getDefaultUncaughtExceptionHandler();
-
     private Piwik(Application application) {
         this.application = application;
     }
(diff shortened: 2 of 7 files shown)


dotsbb:
>I've removed the uncaught exception handler due to concerns described in #28.

I think we should include this feature, at least because of some users, which could wish to migrate their apps from Google analytics to Piwik SDK.
https://developer.android.com/reference/com/google/android/gms/analytics/Tracker.html#enableExceptionReporting(boolean)

d4rken:
I'd rather we re-add it if someone really needs this. I just can not see this being useful in any but the smallest of apps, and those are usually not the apps that do heavy analytics.

Lets compromise, I've readded it as a method within QuickTrack.class and created a new PiwikExceptionHandler.class that contains all the logic separately.

What do you think about the latest commit?

I would still love for someone to create an issue, arguing for needing this in his app.
Maybe someone created an awesome Piwik server-side extension that gives it a full set of Crashtracking tools, but doesn't share it. I would be very much interested in that :smile:.
Does Google Analytics have this?

mattab:
Once we've supported a feature, it's tricky to remove it afterwards in case some users depend on it. Also it can really be useful for some to be able to track exception (while it's far from perfect as you pointed out). 

Can you create an issue with the problems of Exception handling and few screenshots showing how it's not so useful for you yet?

d4rken:
@mattab True, with my last commit to this PR, i added it again but in a (in my opinion) cleaner way, will have to see whether @dotsbb agrees.

I'm currently using ACRA with an app with 1000k+ active users and just now there are 1954 different types of Exceptions in the database. I pruned the database I think 2 month ago.
95% of those exceptions are bugs i can't fix.
We can split those exceptions up into different categories:
* Users screwing with the apk (Modding/Cracking etc).
* Older app versions
* Users screwing with their device ROM
* Bugs in the Android framework

Now imagine we put those 2k different exceptions into the event category of Piwik.
Unless there is a plugin i don't know about, i don't see how this can be feasonably managed within Piwik.
To organise this we would need:
* Ignore or reject exceptions from older app versions
* Mark issues as fixed
* Options to prune Exception events (by version & age)
* Sorting/Filtering by app versions / android versions

To then make the exceptions useful we would also need more information:
* App versions
* ROM details
* Full stacktraces (Crashing tracking without full stacktraces??)
(Compare with ACRA/Crashlytics)

See screenshots, now imagine packing that with 2000 different events of type "Exception" from different app versions and Android versions.
This can work for a very small app, but as soon as you are getting serious numbers, it's too chaotic. 

Piwik should focus on analytics, not try to be a crashtracker.
We can actually put that on a scale.
On the very left as pure crashtrackers we have something like ACRA.
In the middle there are hybrid solutions such as Flurry or Crashlytics.
On the right we have Piwik / GA as pure analytics.

Hm this might have came off a bit as a rant, sorry not my intention :).

d4rken:
Maybe I'm looking at this wrong and the real use case is tracking flows like:
EntryA -> PageA ->PageB->EventC->ExceptionD
Then you could see this as page flow ending as a crash causing you to do further investigation or being able to use this information in conjunction with a real crashtracker.

It seemed to me that the functionality was being "sold" as Bugtracker/Crashtracker alternative, now I'm no longer so sure...

Did i just give myself the legit usecase i was looking for?

In any case the current PR no longer removes the functionality.

mattab:
@d4rken it's great to read your thoughts here. Improving management and display of exceptions is in the scope  to Piwik platform. Would you mind creating an issue eg. "Piwik to measure Errors and Exceptions of apps" or similar? and copy paste there your comments... it would be great to discuss it over there

d4rken:
@mattab 
Here? https://github.com/piwik/piwik-sdk-android/issues
Or there? https://github.com/piwik/piwik/issues

mattab:
I meant there: https://github.com/piwik/piwik/issues :+1:

dotsbb:
@d4rken I like your changes, but I think it's right time to merge them into dev branch. So please change the target of Your PR.

d4rken:
@dotsbb will do.

Issue #23, opened 2015-03-17
Why are Piwik instances tracked against a HashMap of Application objects


d4rken:
AFAIK, the Application object is a singleton that lives as long as the app process exists. There will not be a case where we pass different application objects to Piwik.getInstance(...).

I think we could just remove the HashMap. Did I miss anything?

dotsbb:
You totally right, there is no need of HashMap. It could be rewritten by users (if they want more than 1 tracker) to something like  this https://developers.google.com/analytics/devguides/collection/android/v4/#tracking-methods

d4rken:
This will be resolved by #34

PR #34, opened 2015-03-25
Refactoring for #23 and more

d4rken:
* Simplified getInstance synchronization
* Removed the HashMap<Application,Tracker>, see #23
* ~~Each Tracker now has their own preferences based on apiurl+siteid. So in the future we can track one-shot events per different trackers. This might see some further changes when i submit a PR with changes for download tracking. Note: If you would update the library and call trackDownload again, it would send a new event as the old preference file was per piwik instance not per tracker.~~
* api-url and site-id are now final and can only be set when creating the tracker. It was previously already not possible to create a tracker with no api-url, why would you then want to create a tracker with one url and then change it later. This also prevents concurrency issues, what if the api-url/siteid get changed while some other routines are accessing them. TL;DR Enforcing the practice of creating a new Tracker for a new api-url/siteid.
* Updated all tests to fit the changes and fixed sharedpreferences not working correctly for testcases.

Some autoformatting and import optimization was done too and each time i change a specific variable i rename it to **m**Variable. m is for member, while in the end it's personal preference, it's in line with naming in official AOSP classes, i hope no one is annoyed by that.

Tests all pass on my machine and this is the first PR against the new dev branch :+1:

### DIFF
--- a/piwik_sdk/build.gradle
+++ b/piwik_sdk/build.gradle
@@ -16,7 +16,7 @@ dependencies {
     repositories {
         mavenCentral()
     }
-
+    compile 'com.android.support:support-annotations:21.0.0'
     // Espresso
     androidTestCompile('com.android.support.test.espresso:espresso-core:2.0')
     androidTestCompile('com.android.support.test:testing-support-lib:0.1')
--- a/piwik_sdk/src/main/java/org/piwik/sdk/Piwik.java
+++ b/piwik_sdk/src/main/java/org/piwik/sdk/Piwik.java
@@ -18,8 +18,6 @@ import java.util.HashMap;
 public class Piwik {
     public static final String LOGGER_PREFIX = "PIWIK:";
 
-    private final static Object lock = new Object();
-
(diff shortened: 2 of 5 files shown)


d4rken:
After further thought having different preferences per tracker is something we should leave up to devs and for simple purposes the previous way of one shared preference for Piwik is the best solution.
commit dbe63a3 restored that.

Issues i saw with the reverted approach:
* It's no longer possible to change apiurl/siteid without loosing the previous preferences.
* It might be more interesting to have global piwik file to deal with "opt out" and other possible tracker settings.
* The caveat that using different trackers and the "trackAppDownload" method wouldn't work toegether as it would only fire for the first tracker is not something we should be concerned with now, if someone really needs 3 trackers with different "trackAppDownload" events they could just wrap the "trackNewAppDownload" themself.

dotsbb:
@d4rken Could You merge your branch with actual `piwik:dev`, so I could merge it automatically?

d4rken:
Yeah just saw that, currently fixing the conflicts.

d4rken:
@dotsbb Should be good to go after travis is done :+1:

PR #35, opened 2015-03-26
Refactored test setup

d4rken:
I was trying to fake the application enviroment further, more specifically being able to query the PackageManager for specific information. I needed this to test more detailed download tracking, but the previous architecture didn't allow it.

I then ended up with some major changes, notably:
* We have a custom TestRunner that specifies a custom TestLifeCycle which in turn instantiated our custom Application object.
* There are no more static objects within the TestTracker class, i don't know why this was done before. I'm guessing it was related to the "BeforeClass" method which had to be static.
* We no longer need a "clearTracker()" method because there are no objects that persist across tests. Each test can request new Tracker objects as it likes and reuse them where necessary.

But i think it's easier to understand now, closer to a real app environment and well enhanceable for future tests.

Was there a specific reason for the previous setup and i completely missed something @dotsbb ?

### DIFF
--- a/piwik_sdk/src/test/java/org/piwik/sdk/FullEnvTestLifeCycle.java
+++ b/piwik_sdk/src/test/java/org/piwik/sdk/FullEnvTestLifeCycle.java
@@ -1,0 +1,27 @@
+/*
+ * Android SDK for Piwik
+ *
+ * @link https://github.com/piwik/piwik-android-sdk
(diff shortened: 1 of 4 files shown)


dotsbb:
@d4rken There wasn't any specific reason of doing it in that way, besides of that fact that it was my first try of using Robolectric test suite ;)

d4rken:
What works works :smile:. Do you have a public project with which you are using the sdk?

PR #40, opened 2015-03-28
More detailed download tracking

d4rken:
This is a major enhancement of tracking app downloads.

Previous download stats looked like this:
```
http://packagename/
```
This just allows us to know the app was downloaded, which isn't really a lot.

Now depending on the option it looks like this: exampledownloadstat
```
http://packagename:versioncode/apk-md5-checksum
http://packagename:versioncode/installer-packagename
```
It now not only tells us that the app was downloaded, but also what version of it was downloaded.

* *ExtraIdentifier.INSTALLER_PACKAGENAME* is useful if the same app is deployed on different appstores. Most major appstores set this field (Google/Amazon/Samsung) giving you can overview where your downloads are coming from. Note: This value can be set by tools and the user so some random values will appear in the stats.
* *ExtraIdentifier.APK_CHECKSUM* When your app is in the wild it will sooner or later get changed or "modified", choosing this stat gives you an overview over the differation incarnations there are. Some app stores (e.g. Amazon) modify the apk too. If you match the checksums to it's origin this option becomes even more interesting.

This PR also changes the "trackAppDownload" to only trigger ONCE per VERSION, so after each update it fires once.

The default option with "trackAppDownload" is *INSTALLER_PACKAGENAME*. I do consider the checksum more interesting, but the installer name is "easier accessible" for new developers using this. Additionally the checksum is computational expensive and needs to be done on a thread.

### DIFF
--- a/piwik_sdk/src/main/java/org/piwik/sdk/Tracker.java
+++ b/piwik_sdk/src/main/java/org/piwik/sdk/Tracker.java
@@ -9,10 +9,15 @@ package org.piwik.sdk;
 
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
 import android.util.Log;
 
+import org.piwik.sdk.tools.Checksum;
 import org.piwik.sdk.tools.DeviceHelper;
 
+import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.text.SimpleDateFormat;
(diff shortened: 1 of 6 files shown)


d4rken:
Oh almost forgot, you can technically create ```Context``` objects for apps other than your own and  ```trackNewAppDownload``` takes a context as argument, so you can track other apps too.

d4rken:
@mattab This should be revisited when Piwik further moves towards being an all platform tracker. TAG https://github.com/piwik/piwik/issues/5257 and https://github.com/piwik/piwik/issues/4734

Currently Piwik only accepts valid URLs in the "download" category. This obviously doesn't make much sense here. I imagine the web UI would allow for more hierachy elements and arbitrary strings with some delimiter for hierarchy creation.
Example:
```
packagename/versioncode/extra

packagename/        X+1
    versioncode/       X+1
        extra               X+1
```

Although i think "http://packagename:versioncode/extra" works quite nicely it feels a bit unclean.

d4rken:
track extra app

dotsbb:
:+1: Awesome.

d4rken:
If you don't see any issues, from my side I'm done and this could be merged.

PR #42, opened 2015-04-01
Persist userid

d4rken:
* Removed setUserId that took a *long* paramter, was confusing you can just as well pass the long as string
* Altered the javadocs which stated that the userid is somehow hashed, which isn't the case
* The userid is now persisted in the preferences
* If no userid is set (or persisted) the sdk will create a new userid (random UUID) and persist it
* You can set the userid to NULL to clear, but not to an empty string
* It's no longer possible to have no userid, because why would you want that?


d4rken:
Thanks.

Issue #16, opened 2015-02-10
Offline mode missing


Twibit:
If the [dispach](https://github.com/piwik/piwik-sdk-android/blob/7d6df25bd68dbcb1799c93d36bcde904a0c9e42a/piwik_sdk/src/main/java/org/piwik/sdk/Tracker.java#L134) of  events occurs when the device is offline, the events queue is cleared before knowing if the request was a success.

d4rken:
I'm probably going to tackle this before going into production with piwik in my Android app. Any ideas on a solution direction?

dotsbb:
I think there should be some FIFO container within Tracker instance so we could store here untracked views. It would be great if this container would be configurable (for example it might accept `MAX_SIZE` param).

d4rken:
A few points that we should consider:

* We should store somekind of version code or error handling with the offline data. Future updates might break compatibility with previously stored but unsent offline data. Best case, still reuse it, in any othercase discard it.
* The local storage should be limited to a configurable size. We could add a setting that allows the dev to choose betwen storing in /files or  /cache. Within cache files may be removed automatically if space runs low, but we might loose offline data if the user chooses to clear cache.
* The local storage should probably be a buffer such that if we exceed the limit, the oldest offline data is overwritten.
* As we are now dealing with transmission failures, there should be some kind of system that deals with "retries".
* An analytics framework should minimize it's impact on the app and device, so if we regain connectivity, and there is a lot of offline data, we shouldn't send it all at once, but gradually transmit the views. This will also limit the impact this has on the server.
* If the piwik instance is offline, this would count as transmission failure. If we have a million app users, and our piwik instance goes offline, and then goes online again, all users transmitting all offline data would probably DDOS the piwik instance.
* We could gradually transmit the offline data over a time span. Would we need to store new events if there still is offline or can we send them directly? (e.g. does the server get confused if they are totally out of order)

dotsbb:
@d4rken Have You seen [this fork](https://github.com/piwik/piwik-sdk-android/compare/dev...Bookvideo:fork/bookvideo?expand=1)? It does 0 of yours considered points at this moment but it can be something we can refactor and introduce features incrementally to it.

d4rken:
I have now :wink: . It's one way to solve it but as you mentioned, it addresses none of the concerns. I would probably start from scratch, not refactor this.

[...]

PR #46, opened 2015-04-05
Tracker v2.1 - Harder, Better, Faster, Stronger, Threadsafe

d4rken:
(I somehow broke the previous pullrequest)

The base goal was to make the Tracker class thread safe.

Previously the tracker was the sole responsible entity for the queryparamters.
Multiple threads using the same tracker would overwrite each others values.
Also multiple trackers were no solution as each tracker would automatically start a new session.

The idea is to decouple the *Tracker* from holding specific query data.
It should know where to deliver the queries and supplement basic information if necessary, but not more.
Each thread/request/actor constructs it's own query then passes it to the Tracker.
The Tracker will inspect the query, fill in missing information and then give it to the *Dispatcher* to be dispatched.

Change notes:

* A new tracking session starts after the first track, not on *Tracker* creation. Because we can always reuse the same *Tracker*, all events stay in the same visitor session.
* The base method for all tracking is *getTracker().track(TrackMe)* it takes a *TrackMe* object.
* The *TrackMe* object is a wrapper for the *QueryParams* and the screen scope custom user variables. Google Analytics has a similar *HitBuilder* object.
* All other methods like trackScreen/trackEvent just create a *TrackMe* object and set the correct parameters. I.e. they are convinience methods.
* The *track(TrackMe)* method will insert basic information such as screen resolution and useragent only on the first track of the visit (if not already set). While a dev can force set these values, they normally don't change and Piwik just needs them to be transmitted once per visit session.
* The *track(TrackMe)* method will for each visit insert the basic parameters such visitorId/userId etc. and visitor scope variables.
* The *track(TrackMe)* method will then call "build()" on the *TrackMe* object which creates a String with the query.
* This query is then given to the *Dispatcher*, which is a threadsafe queue with a single worker thread.
* The *Dispatcher* will at each interval drain the queue, transmit the events, then go to sleep again.

* Having to set the screen custom variables before tracking a screen is no longer a possible mistake, because you have to create and set them on a *TrackMe* object before you can pass it.

* All time values are now in milliseconds, because everything is always in milliseconds in Android. Having to check wether it is seconds or milliseconds just creates more errors. Dealing with developers not users, milliseconds are perfectly fine.

* There exists a default *TrackMe* object now which is populated on Tracker creation with all necessary parameters. Before each TrackMe object gets send, the Tracker will try to fill in missing parameters from the default TrackMe, but it won't overwrite any existing values. An advanced developer can still use *getDefaultTrackMe* and modify the default values to achieve anykind of customization.

* Methods such as *setResolution* or *setUserAgent* etc. have been removed as the developer is not encouraged to alter them. But they can still be changed through *getDefaultTrackMe*

* A few freedoms are given back to developers now. If a value gets explicitly NULLed in the default TrackMe object, the Tracker will no longer refill it.

* The tests have been adapted to all changes

* A new testclass *TestDispatcher* now exists that does multithreading & concurrency tests.

General Notes:
* Because the *Dispatcher* is now the sole exit point for all queries to be send, it would be a good target to look at for implementing offline mode (#16).

````

D/launchTestThreads: Launching 100 threads, 1000 queries each
D/launchTestThreads: All launched.
D/checkForMIAs: 5216 events created, 331 requests dispatched, containing 4821 flattened queries
D/checkForMIAs: 10923 events created, 705 requests dispatched, containing 10474 flattened queries
D/checkForMIAs: 16695 events created, 1081 requests dispatched, containing 16116 flattened queries
D/checkForMIAs: 22372 events created, 1449 requests dispatched, containing 21975 flattened queries
D/checkForMIAs: 28038 events created, 1817 requests dispatched, containing 27605 flattened queries
D/checkForMIAs: 33916 events created, 2201 requests dispatched, containing 33287 flattened queries
D/checkForMIAs: 39640 events created, 2571 requests dispatched, containing 39158 flattened queries
D/checkForMIAs: 45351 events created, 2945 requests dispatched, containing 44889 flattened queries
D/checkForMIAs: 51173 events created, 3317 requests dispatched, containing 50605 flattened queries
D/checkForMIAs: 56970 events created, 3692 requests dispatched, containing 56368 flattened queries
D/checkForMIAs: 63157 events created, 4090 requests dispatched, containing 62234 flattened queries
D/checkForMIAs: 69110 events created, 4483 requests dispatched, containing 68397 flattened queries
D/checkForMIAs: 75253 events created, 4886 requests dispatched, containing 74400 flattened queries
D/checkForMIAs: 81301 events created, 5278 requests dispatched, containing 80486 flattened queries
D/checkForMIAs: 87529 events created, 5681 requests dispatched, containing 86590 flattened queries
D/checkForMIAs: 93659 events created, 6074 requests dispatched, containing 92723 flattened queries
D/checkForMIAs: 99199 events created, 6450 requests dispatched, containing 98519 flattened queries
D/checkForMIAs: 100000 events created, 6618 requests dispatched, containing 99991 flattened queries
D/checkForMIAs: 100000 events created, 6618 requests dispatched, containing 100000 flattened queries
D/checkForMIAs: All send queries are accounted for.
````
example

### CODE REVIEW

**Comment by d4rken** [comment]
I noticed Piwik does accept http, https and ftp, so I changed the restriction to let through ftp protocol based URLs too. Added a wrapper function for tracking outlinks *trackOutlink(URL)* + tests.

File: piwik_sdk/src/main/java/org/piwik/sdk/Dispatcher.java [hunk 1]
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    private Runnable mLoop = new Runnable() {
+        @Override
+        public void run() {
+            android.os.Process.setThreadPriority(Process.THREAD_PRIORITY_BACKGROUND);
+            while (mRunning) {
+                try {
+                    // Either we wait the interval or forceDispatch() granted us one free pass
+                    mSleepToken.tryAcquire(mDispatchInterval, TimeUnit.MILLISECONDS);
```
dotsbb: Very clever :+1:

piwik_sdk/src/main/java/org/piwik/sdk/Dispatcher.java
```diff
@@ -0,0 +1,252 @@
...
+    private boolean doRequest(HttpRequestBase requestBase) {
+        HttpClient client = new DefaultHttpClient();
+        HttpConnectionParams.setConnectionTimeout(client.getParams(), mTimeOut);
+        HttpResponse response;
+
+        if (mPiwik.isDryRun()) {
+            Logy.d(LOGGER_TAG, "DryRun, stored HttpRequest, now " + mDryRunOutput.size());
+            mDryRunOutput.add(requestBase);
+        } else {
+            mDryRunOutput.clear();
```
dotsbb: I'll rather check if `mDryRUnOutput` is not empty before trying clearing it (mainly bc of [this](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/ArrayList.java#473))

d4rken: It would be a bit more efficient, but I don't think it's noticeable. This happens once per dispatch intervall or once per batched page, i.e. with 2-3digit miliseconds wait times.

I still agree though and will submit another PR for this, but I'd prefer to have this PR merged first instead of further extending this one though.

Or do you mean something different than the efficiency?

piwik_sdk/src/main/java/org/piwik/sdk/TrackMe.java
```diff
@@ -0,0 +1,131 @@
...
+     * set(QueryParams.HOURS, "10");
+     * set(QueryParams.MINUTES, "45");
+     * set(QueryParams.SECONDS, "30");
+     * </pre>
+     *
+     * @param key   query params name
+     * @param value value
+     * @return tracker instance
+     */
+    public synchronized TrackMe set(@NonNull QueryParams key, String value) {
```
dotsbb: There is any reason why all `TrackMe` public methods are synchronized?

dotsbb:
Thank You for such a great contribution!

piwik_sdk/src/main/java/org/piwik/sdk/TrackMe.java
d4rken: It's an improvement, but not a solution.

Specifically the case of a developer handeling the TrackMe object between two threads, and the developer modifying the TrackMe object after passing it to the Tracker are problems.

Synchronizing the methods against the class was my idea to add some protection against that.
* While doing trySet it will never overwrite a value.
* While doing build the custom variables can not change.
* HashMap is not thread-safe
* CustomVariables are not threadsafe, we don't want them changing when putting them into the params in build()

But I'm not sure now how good this solution is though now you mentioned it.... The developer should not get ideas to try stuff like that anyways ;). If those methods can throw ConcurrencyModificationExceptions, they now can't.
We could go for a ConcurrentHashMap and less locking on the methods as potential improvement, while also making CustomVariables thread-safe.
Performance wise the synchronize statements can be neglected as TrackMe should not be accessed by multiple threads, but i agree that it's not a very clean approach.

d4rken:
@dotsbb 
It wasn't completely altruistic, I needed thread-safety to track events that come from multiple threads in my app :wink: . It's currently in public beta test. Looking good so far.
I also had a little bit fun implementing this :smile:.

Also thank you for all your work! I wouldn't have had the motiviation to do this from scratch. You are not affiliated with Piwik, are you?

Next target would be #16 , don't have a good idea yet though.

dotsbb:
@d4rken I'm a python developer on a daily basis, and don't have a deep knowledge of Piwik at all. Let's say this SDK is side effect of 10 days "hackathon" where I was only participant ;)

Issue #18, opened 2015-02-18
User visits do not seem to be tracked properly.


sdunnin:
We are registering visits from an Android application. When viewing from the dashboard visits and unique visitors are accurately reflected, however, when calling API method VisitsSummary.getUsers and providing parameter &segment=visitorType==returning the value returned is always zero. The user visits all register as new users even though the users have multiple visits across several days. 

Website visits reported via the javascript client do report new vs. returning visitors. Both Javascript and Java clients are sending the userid via the uid paramater.

d4rken:
Are you storing a user-id?
As i currently understand it the visitor-id is generated and not persisted so will only track the user for his current visit.
Unless you persisted a user-id which is reused each time the app is open, piwik will not be able to track the user accross multiple visits.

sdunnin:
Yes, we pass a consistent userid in the api call from the Android application.

d4rken:
I don't think this issue will see any progress without further testing from your side.
I suggest to debug your app to make sure that a userID is being transmitted.

If you can not find anything through debugging try to create a simple as possible app that reproduces this.

Possibly you spot an issue and we can enhance the unit tests to catch this in the future.

I'm currently still integrating the sdk into my app and if i find anything I'll report back.

d4rken:
In my live test the Piwik webpage seems to report a non-zero value for returning visists, although this is with the current dev-branch i found no issues during my changes that would suggest an issue with this.

> when calling API method VisitsSummary.getUsers and providing parameter &segment=visitorType==returning the value returned is always zero.

As the API method for this is not part of the Android-SDK but the main Piwik SDK wouldn't it make sense to close this ticket and create one here: https://github.com/piwik/piwik/issues

d4rken:
I think we need to reopen this @dotsbb 

I can randomly pick entries from the live visitor log and see that they have multiple visists across multiple days. "Visits by days since last visit" aswell as "Visits by Visit Number" should have significantly higher values in other rows.

![visits](https://cloud.githubusercontent.com/assets/1439229/7109363/b41f6760-e19c-11e4-8975-a4fb6d32303f.PNG)

As @sdunnin issue was reported before the Tracker rewrite, this doesn't seem to be a regression from my Tracker rewrite.

* Creating a segment for "Visitor Type IS NOT" only offers autocomplete to "NEW" and shows a too low number
* Creating a segment for "Number of visits AT LEAST 2" shows 0 zero results.
[...]

Issue #14, opened 2015-01-16
Add SNI support


gilbert1991:
Hi, when I test on my piwik sever, I met error like below:
javax.net.ssl.SSLException: hostname in certificate didn't match: analytics.test.com != test

dotsbb:
As far as I know Apache HTTP Client used in Piwik SDK doesn't support Server Name Indication.
Try to use a distinct IP address (so as not to need [SNI](http://en.wikipedia.org/wiki/Server_Name_Indication)).

Meanwhile I'll try to change  HTTP Client to  new port of [HttpClient for Android](https://hc.apache.org/httpcomponents-client-4.3.x/android-port.html) which support SNI and TLS v1.1/v1.2.

gilbert1991:
I think it is the case I found on Stackoverflow: http://stackoverflow.com/questions/21956663/why-does-android-get-the-wrong-ssl-certificate-two-domains-one-server. Reference for others.

d4rken:
From https://developer.android.com/training/articles/security-ssl.html
>Fortunately, HttpsURLConnection supports SNI since Android 2.3. Unfortunately, Apache HTTP Client does not, which is one of the many reasons we discourage its use. One workaround if you need to support Android 2.2 (and older) or Apache HTTP Client is to set up an alternative virtual host on a unique port so that it's unambiguous which server certificate to return.

If we raise the minAPI to at least 8 (currently 7) this looks doeable without requiring extra 3rd party libs.
If we look at user numbers in the wild, going minAPI 8 seems to no big deal.

dotsbb:
@d4rken good to hear. I'll vote to bump minAPI to the 8 level.
According to this [chart](http://www.appbrain.com/stats/top-android-sdk-versions) 2.0-2.1 Eclair hits only 0.1% of users (testers perhaps? ;) )

d4rken:
+1 from me too.
![top10api-tools](https://cloud.githubusercontent.com/assets/1439229/7055249/5ee4a0ec-de42-11e4-8149-1fab957d0e31.PNG)

PR #50, opened 2015-04-14
Feature: Visit count, frequency, interval

d4rken:
This PR addresses #18 by implementing the following parameters:

* "_idvc" aka Total-Visit-Count
* "_viewts" aka Previous-Visit-Timestamp
* "_idts" aka First-Visit-Timestamp

The Visitors -> Engangement page should now be correctly populated.
Segmentation for returning visitors should now also work.

Notes:
* The parameters are initialized/updated on Tracker creation not on first track. This prevents concurrence issues and deals with the fact that each Tracker has it's own visit session.
* The visit count is modified even when the user has chosen to opt-out. If he opts in at a later point in time, accurate values are transmitted.
* This PR increases the min API to 10 (Gingerbread 2.3.3+) to make use of SharedPreferences.apply() which improves performance, and in regards to #14.
* Similar to other one-time parameters these too are only transmitted on the first query per visit to conserve bandwidth.

### DIFF
--- a/piwik_sdk/build.gradle
+++ b/piwik_sdk/build.gradle
@@ -9,7 +9,7 @@ android {
     buildToolsVersion '22.0.1'
 
     defaultConfig {
-        minSdkVersion 7
+        minSdkVersion 10
         targetSdkVersion 22
         versionCode 1
         versionName '0.0.1'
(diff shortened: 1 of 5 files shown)


dotsbb:
Thanks, @d4rken!

PR #51, opened 2015-04-14
Feature: Smart dispatch

d4rken:
From the Piwik application object we can get callbacks for memory events:
* Dispatch queued events when the app UI goes into the background (e.g. user is done with the app and presses the home button)
* Dispatch queued events when the app is running the risk of being killed to free memory. This makes sure that queued events are not lost and by emptying the queue we also free memory. Win-Win :smile: 

Side note:
* As Tracker is now threadsafe the Piwik application class was cleaned up a bit. We don't want to advertise using authToken and the develoepr should be encouraged to reuse the Tracker object.


dotsbb:
@d4rken You rock!

d4rken:
:blush:

PR #53, opened 2015-04-20
Bugfixes: Visit time values & Race-condition on session start

d4rken:
This PR fixes two major issues.

The first issue is that during my previous implementation of previous-visit and first-visit i overlooked that Piwik expects the timestamps in seconds (unix timestamp is always in seconds, i overlooked that) and used **System.currentTimeMillis()** ... *sigh* This is fixed now by just dividing the value through 1000.
We now also have tests for this.

The second issue is a race-condition on session start.
The first thread that tracks something will do the session-start initialization, i.e. information that only needs to be transmitted once. While loading/inserting these values into the TrackMe request, it is possible that a second thread overtakes the first one and puts his transmission (missing the initial transmission infos) into the Dispatcher queue before the first thread does so.
I've made changes to fix this. A CountDownLatch now prevents other threads from dispatching events until the first thread put his "new-session" event into the Dispatcher queue.
We now also have tests for this.

### DIFF
--- a/piwik_sdk/src/main/java/org/piwik/sdk/Dispatcher.java
+++ b/piwik_sdk/src/main/java/org/piwik/sdk/Dispatcher.java
@@ -72,6 +72,10 @@ public class Dispatcher {
         mAuthToken = authToken;
     }
 
+    /**
+     * Connection timeout in miliseconds
+     * @return
+     */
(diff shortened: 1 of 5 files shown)


dotsbb:
Didn't know about `CountDownLatch`, great work as usual.

PR #56, opened 2015-08-04
Version 0.0.2

dotsbb:
**Change log**

* Android  API level increased to 10 (Gingerbread 2.3.3+) 
* The `opt-out` and `userid` vars are now persisted in the preferences
* Referrer tracking on `trackDownload`
* Feature: Visit count, frequency, interval
* Threadsafe and faster tracker core.
* Smarter dispatching (on low memory)

**Bugfixes**
* Visit time values
* Race-condition on session start
* `trackAppDownload` bug


d4rken:
Oh wow, it didn't feel like so many commits when working on it :smile: 
Looks good to me :+1: , I've been using the dev branch in production for quite some time.

Issue #31, opened 2015-03-25
Proposal: Remove compiled jar from repo, attach to tagged releases


d4rken:
Currently the repo contains the compiled jar here:
https://github.com/piwik/piwik-sdk-android/tree/master/piwik_sdk/jar
It's difficult to track against which code version it was built and everytime you make a testbuild it gets changed and suggested to be recommitted by git.
Additionally it's quickly outdated and we don't want (do we?) have to build and update it with each PR.
(Currently the jar doesn't already match the latest code).

I would suggest to remove the jar from the repo and attach it to each release as download:
https://github.com/piwik/piwik-sdk-android/releases

@dotsbb or @mattab could then choose certain commits, tag it and create a new release.

mattab:
or maybe we could build the JAR from travis so it's done automatically on each commit?

d4rken:
If that is possible that would be cool :).
We should differentiate between some "stable" and one "bleeding edge" version though.

mattab:
Each commit has its own build. So then when you make a release on github it will create a tags/x.y.z and a commit, which will trigger a build which could build the Android SDK.

PR #57, opened 2015-08-16
Publish on jCenter

### LINKED ISSUE #17: Publish to Maven
It would be great if this SDK could be available as a maven/gradle dependency.

Maybe you could use [gradle-mvn-push](https://github.com/chrisbanes/gradle-mvn-push) to do it.

### LINKED ISSUE #55: Merge develop into master.
@dotsbb I think we can merge the current develop branch into master.
I've been using it in production for a month now and I'm satisfied.

New people looking to integrate Piwik into their app shouldn't use the build from current master branch.

We could increase the version to 0.0.2 as there were some major changes.

When merging we should also make sure the .jar file is either updated or removed to prevent confusion.

Not sure whats going on with #31

d4rken:
I've build the latest develop branch and submitted it to jCenter, which replaced Maven as the default repo for new Android projects.
Within the next few hours it should become available under:
```
compile 'org.piwik.sdk:piwik-sdk:0.0.2'
```
This PR updates gradle and adds scripts used for uploading to bintray.com which hosts jCenter.
The package is uploaded under my jCenter account, so I have to update the package when it's necessary, but I don't think this will be an issue.
And the README has also been updated to reflect the new (easier) install instructions.

@dotsbb 
After merging this we can close #17, and also #31 because it is then obsolete. Then you could merge your PR #56 then close #55 and everything is tidy again :smile:


dotsbb:
Nice job, Matthias :+1: Merging

Issue #27, opened 2015-03-23
Proposal: Get rid of Piwik/Tracker differentation


d4rken:
I see no need for the Piwik AND Tracker class except to be able to have multiple Trackers for thread safety.
What if we make the Tracker abstract & threadsafe.
e.g.
```
class MyTracker extends PiwikTracker {
    @Override
    public String getTrackerUrl() {
        return "foo";
    }
}
```
Then in your code you just call MyTracker.get(getContext()).doStuff()

* The HashMap approach in Piwik.class already seems superfluos, see #23
* Dry run / opt out can be set directly.
* autoBindActivities is currently not working correctly anyways and can be moved into an extra class if we really want it, but then go along something like Tracker.bind(Activity), we could also look into Tracker.bind(Fragment)

Would love some comments on this.

dotsbb:
I believe that managing `Tracker` instances should be similar to the [Google Analytics SDK Tracker](https://developers.google.com/analytics/devguides/collection/android/v4/). @mattab please confirm if it's true.

`autoBindActivities` is not mandatory and could be refactored.

d4rken:
I understand, i didn't know that staying close to GA was desired.
Or maybe i could have figured that as Piwik itself is quite similar.

I'll submit individual PR for the other changes i mentioned, excluding this Piwik/Tracker merge.

I'll look into adding a helper class to the SDK that would allow it to stay close to the GAnalytics design but still offer a few shortcuts to make it easier to use.

Hearing @mattab option about this would be cool though.

dotsbb:
@d4rken Perfect! I look forward to your PRs.

mattab:
quick feedback: I'm not familiar with Android development so unfortunately I don't know much more. but the general idea is that if Google implemented it this way for their SDK, I think there would be a reason for it, maybe some complex use case that would benefit from this feature...?

d4rken:
@mattab but there is no direct goal to stay close to their api design? looking at other APIs, e.g. In-App-Purchases API, they sometimes make it overly complicated.

In any case the documentation linked by @dotsbb shows a similar approach to as our SDK currently uses just sans the HashMap, while mentioned in Googles docs, it's an example how the dev could implement it, it's not there by default.

I agree that we should keep it this way so as to leave a dev the option to extend it the same way that Google proposes it in their documentation.

[...]

PR #59, opened 2015-09-01
Ecommerce

lkogler:
- Added ecommerce support based on the javascript API: http://developer.piwik.org/api-reference/tracking-javascript#ecommerce.

- Fixed flaky test that depended on order of iteration through a HashMap


d4rken:
Thanks for the contribution!

Some thoughts:
* Prefixing global variables and running Android Studio's code formatter (saw some extra spaces).

*  Do we need to worry about what type of currency the user is using?

*  I don't think we should have ```EcommerceItems``` as global variable in ```Tracker```. The goal is to decouple the Tracker as holder of specific event data (compare #46 and before). I find this is cleaner and we also don't have to worry about multithreading.
The Tracker only has the responsibility to fill in missing values and put forward the request.

What about extending extending ```TrackMe``` to something like ```EcommerceTrackMe``` which then offers convinience functions to add/remove items. The developer could persist this ```EcommerceTrackMe``` object within the app in any way desired and then reuse it. What do you think?

Side note: I think we should then also make TrackMe and it's extension ```Parcleable``` to help devs a bit.

piwik-sdk/src/main/java/org/piwik/sdk/ecommerce/EcommerceItems.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class EcommerceItems {
+    public Map<String, List<String>> items = new HashMap<>();
+
+    public void addItem(String sku, String name, String category, int price, int quantity) {
+        List<String> item = Arrays.asList(sku, name, category, CurrencyFormatter.priceString(price), "" + quantity);
+        items.put(sku, item);
+    }
+
+    public String toJson() {
```
d4rken: Better use JSON classes for this, see:
https://github.com/piwik/piwik-sdk-android/blob/dev/piwik-sdk/src/main/java/org/piwik/sdk/CustomVariables.java

lkogler:
- Sure, I can run the Android Studio formatter

- As far as I can tell, only dollars are supported on the server side (so far), but correct me if I'm wrong.  That seems like a great thing to support in the future, but is outside the scope of what I would like to do for this pull request.

- My original thought was actually to have the user be responsible for creating the `EcommerceItems` object and passing it to the `trackEcommerceOrder` and `trackEcommerceCartUpdate` functions as a parameter.  The reason I went with this design was to mirror the existing javascript API; however, if that is not a concern, I'm fine pulling it out!

d4rken:
Yeah you are right with the currencies, i was under the impression the device locales played into it, but they don't.

I mirrored my previous changes according to what the [HTTP API docs](http://developer.piwik.org/api-reference/tracking-api) describe and after Google Analytics.
[GA old](https://developers.google.com/analytics/devguides/collection/android/v4/ecommerce) actually had an extra tracker type for ecommerce, but this is deprecated now. Their [new setup](https://developers.google.com/analytics/devguides/collection/android/v4/enhanced-ecommerce) uses a similar approach to what i described, passing specific ```Product``` objects to a single tracker.

For me the Javascript API is of no concern, but this is not "my" project. @mattab and @dotsbb would have to chime in :smile: . When this previously came up though in #27 staying to close to GA was good.

dotsbb:
@lkogler thanks for Your work!

I will agree with @d4rken -  we shouldn't have `EcommerceItems` as global variable in `Tracker`. 
There was a lot of work done by @d4rken in order to ensure thread-safe in SDK.

We would like to keep SDK API as close to GA as possible. 
Maybe `trackEcommerceOrder` and `trackEcommerceCartUpdate` methods might accept `Product` or `EcommerceItems` as params. 
Also I like the `Parcleable` / `EcommerceTrackMe` idea.

lkogler:
Ok, I removed the ecommerceItems global variable from the Tracker class.  It must now be passed as a parameter.  

Also cleaned up the code a bit.  Look good?

piwik-sdk/src/main/java/org/piwik/sdk/ecommerce/EcommerceItems.java
```diff
@@ -0,0 +1,79 @@
...
+import org.json.JSONArray;
+import org.piwik.sdk.tools.CurrencyFormatter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class EcommerceItems {
+    public Map<String, JSONArray> items = new HashMap<>();
+
+
```
d4rken: Not an issue, just FYI, with prefixing i meant "items" -> "mItems". Code formatting looks good now.

d4rken:
Rest looks good, also nice work on the unit tests.

There is probably a good reason for it, but why is currency being passed in cents instead of a float if we later convert it to (string) float anyways?

I still like the idea of an ```EcommerceTrackMe```, but we could merge this into dev and create a HelpWanted ticket for it.

lkogler:
It is generally considered bad practice to use floating point numbers for currency calculations (http://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency).  I would be OK with having the user pass in a string, but it seemed like there would be less opportunity for user formatting error if we have them pass an integer.

Maybe I'm missing something, but I don't see what value is added by having an `EcommerceTrackMe` object, and it seems like significant additional complexity.  The existing interface is working great for my use case, anyhow ;)

d4rken:
I understand, better keep it in cents then.

Regarding the ```EcommerceTrackMe```, i just like the idea of basing extra functionality on the ```TrackMe``` object instead of the ```Tracker```. Seeing the TrackMe as some kind of arbitrary message format getting extra (header like) information from the Tracker which passes it to lower layer (the dispatcher). But this is just nitpicking now. Maybe I'll submit a PR at some point, maybe not.

:+1: for a merge, code is good, tests are there, functionality is in line with the other methods offered by ```Tracker```. If you have already tested it in your use-case, even better.

dotsbb:
:+1: I'm going to merge it to master soon.

d4rken:
@dotsbb Don't merge it into master though unless the version is raised and it's tagged as v0.0.3. The idea of master/dev branch is to collect PRs/changesets in dev and then merge release batches.

I.e. Master should contain v0.0.2 (also see #60) and dev should have a raised version of v0.0.3, thus calling the bleeding edge version v0.0.3-SNAPSHOT, until we merge that into master, at which point v0.0.3 would be "complete" and no longer a snapshot. Then we raise to v0.0.4 on dev... rinse repeat...

d4rken:
I'll get snapshot builds for the dev branch going and then submit a PR to update the docs.

PR #62, opened 2015-09-24
Ecommerce

dotsbb:
- [x] Added ecommerce support based on the javascript API: http://developer.piwik.org/api-reference/tracking-javascript#ecommerce.
- [x] Fixed flaky test that depended on order of iteration through a HashMap
- [x] Docs


d4rken:
Change the version in the sdk gradle file to v0.0.3 in this PR request, then we should tag the last master commit after merging the PR with v0.0.3.
So this would then be a new release.

We should also make a commit to dev that raises its version to v0.0.4 (e.g. next version in development).
(We neglected to do this the last time).

d4rken:
Merge is completed, I've updated the bintray repo and tagged the release.

Issue #70, opened 2015-12-02
Let users track Custom Dimensions 


tsteur:
In https://github.com/piwik/piwik/issues/9129 we created a new plugin to track Custom Dimensions.

To the JavaScript Tracker we added a new method `setCustomDimension(int dimensionId, string dimensionValue)`. This sets a tracking API parameter `dimension$dimensionId=$dimensionValue`. Eg `dimension1=foo` or `dimension2=bar`. So the tracking API parameter starts with `dimension` followed by the set `dimensionId`. 

I reckon we should be consistent here with the PHP Tracker and clear all set custom dimensions after the next tracking request was sent. 
eg
```
setCustomDimension(1, 'myvalue1')
setCustomDimension(2, 'myvalue2)
trackPageview()
// now dimension 1 and 2 would be cleared and one has to set values again
```

FYI: Custom Dimensions are similar to Custom Variables. Currently a plugin is needed to use them but in Piwik 3.0 it will be installed by default and at some point we will deprecate Custom Variables (not soon though)

tsteur:
BTW: In PHP-Tracker we used a more generic implementation see: https://github.com/piwik/piwik-php-tracker/pull/15/files 

It's basically a method `setCustomTrackingParameter(parameter, value)` that let's you set a dimension like this: `setCustomTrackingParameter('dimension1', 'my value')`. Also in the future it will let you set any custom tracking parameter for any custom plugin.

d4rken:
I favor the generic approach especially as the functionality is still a plugin, additionally this would be flexible for future enhancements.

We could note the planned changes for ```setCustomVariables``` (maybe @deprecate ?) in the javadocs and add the plugin as example to the javadoc for ```setCustomTrackingParameter```

For future reference: https://github.com/piwik/plugin-CustomDimensions

dotsbb:
@tsteur @d4rken I'll do this one using generic approach.

tsteur:
> @tsteur @d4rken I'll do this one using generic approach.
:+1: 

> We could note the planned changes for setCustomVariables (maybe @deprecate ?)

They are not deprecated yet, maybe in 1-3 years. No need to add it yet but later for sure

PR #71, opened 2015-12-04
Custom Dimensions

dotsbb:
Ref #70

### DIFF
--- a/piwik-sdk/src/main/java/org/piwik/sdk/plugins/CustomDimensions.java
+++ b/piwik-sdk/src/main/java/org/piwik/sdk/plugins/CustomDimensions.java
@@ -1,0 +1,35 @@
+package org.piwik.sdk.plugins;
+
+import org.piwik.sdk.Piwik;
+import org.piwik.sdk.tools.Logy;
+import org.piwik.sdk.TrackMe;
+
+/**
+ * This plugins allows you to track any Custom Dimensions.
+ * In order to use this functionality install and configure
+ * https://plugins.piwik.org/CustomDimensions plugin.
+ */
+public class CustomDimensions extends TrackMe {
+    protected static final String LOGGER_TAG = Piwik.LOGGER_PREFIX + "CustomDimensions";
+    /**
+     * This method sets a tracking API parameter dimension%dimensionId%=%dimensionValue%.
+     * Eg dimension1=foo or dimension2=bar.
+     * So the tracking API parameter starts with dimension followed by the set dimensionId.
+     * @param dimensionId  accepts values greater than 0
+     * @param dimensionValue is limited to 255 characters
+     * @return instance of CustomDimensions plugin
+     */
+    public synchronized CustomDimensions set(int dimensionId, String dimensionValue) {
+        if (dimensionId < 1){
+            Logy.w(LOGGER_TAG, "dimensionId should be great than 0");
+            return this;
+        }
+        if (dimensionValue != null && dimensionValue.length() > 255){
+            Logy.w(LOGGER_TAG, "dimensionValue will be truncated to 255 chars");
+            dimensionValue = dimensionValue.substring(0, 255);
+        }
+        set("dimension" + dimensionId, dimensionValue);
+        return this;
+    }
+
+}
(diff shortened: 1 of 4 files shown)


d4rken:
Maybe some javadoc to say that customvars are being deprecated at some point and some javadoc to say currently customdimensions require a plugin (insert plugin link).

dotsbb:
@d4rken Sure, I'll update javadoc but I can't figure out how we can replace `CustomVariables`  with new plugin aproach 	though. Some users may want to sent *customvars* **and** *custom dimensions* in one request.

d4rken:
AFAIK the new plugin approach is not there to replace the current approach. It's a switch that will be done with Piwik 3.0 and for now requires an extra plugin. So currently you can install the plugin and do both customvariables and customdimensions.

d4rken:
The current custom variables approach does [json conversion](https://github.com/piwik/piwik-sdk-android/blob/dev/piwik-sdk/src/main/java/org/piwik/sdk/CustomVariables.java), so we can't really replace it anyways as the server wouldn't understand the syntax ```dimension$dimensionId=$dimensionValue``` without the plugin and even then, would it group the information in the same way? Probably not.

piwik-sdk/src/main/java/org/piwik/sdk/plugins/CustomDimensions.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+public class CustomDimensions extends TrackMe {
+
+    /**
+     * This method sets a tracking API parameter dimension%dimensionId%=%dimensionValue%.
+     * Eg dimension1=foo or dimension2=bar.
+     * So the tracking API parameter starts with dimension followed by the set dimensionId.
+     * @return instance of CustomDimensions plugin
+     */
+    public synchronized CustomDimensions set(int dimensionId, String dimensionValue) {
```
d4rken: We should probably document the limitations here. See: https://github.com/piwik/plugin-CustomDimensions/issues/6

Actually we could enforce them.
* Check that the ```dimensionId```is valid. I guess it's 0-* or 1-*?
* ```dimensionValue``` is max 255 chars.

Sp if the paramaters are bad we could throw a warning and just not set them.

piwik-sdk/src/test/java/org/piwik/sdk/TrackerTest.java
```diff
@@ -808,6 +809,19 @@ public void testPreviousVisit() throws Exception {
         assertEquals(1000l, Long.parseLong(queryParams.get(QueryParams.PREVIOUS_VISIT_TIMESTAMP)));
     }
 
+    @Test
+    public void testSetCustomDimensions() throws Exception {
+        CustomDimensions customDimensions = new CustomDimensions();
+        customDimensions.set(1, "foo");
+        customDimensions.set(2, "bar");
+        customDimensions.set(3, "empty").set(3, null);
```
d4rken: Good check!
What is the expected behavior client/server side for an empty string ```""``` ?

piwik-sdk/src/main/java/org/piwik/sdk/plugins/CustomDimensions.java
dotsbb: Good spotted.
`dimensionId` is 1+ https://github.com/piwik/plugin-CustomDimensions/blob/master/Dao/Configuration.php#L152

piwik-sdk/src/test/java/org/piwik/sdk/TrackerTest.java
dotsbb: An empty string behaves the same way as `null` value
https://github.com/piwik/plugin-CustomDimensions/blob/e330cb2c0db08d1a5b4e489d3fba7acc0e7cc43c/Tracker/CustomDimensionsRequestProcessor.php#L100

d4rken:
Did you remove the javadoc changes to ```setScreenCustomVariable```? I think we could still encourage people to look at CustomDimensions. As deprecated might be too harsh, maybe something like: 

> This will be replaced by https://plugins.piwik.org/CustomDimensions at some point.

Otherwise, good job,  :+1:  for merge into dev.

btw what does the label CR stand for?

dotsbb:
I have moved javadoc with `setScreenCustomVariable` straight to the master.
CR means that a Code Review is needed.

PR #73, opened 2015-12-05
Coverage bage

dotsbb:


d4rken:
:+1:  nice.

I'm always unsure about testing of private methods.
Have you worked with ```@VisibleForTesting``` ?
Also: http://programmers.stackexchange.com/questions/100959/how-do-you-unit-test-private-methods

dotsbb:
I haven't used `@VisibleForTesting`. I'm not very happy about testing private method too.

PR #83, opened 2016-01-26
Referrer improvements

d4rken:
So this is a new idea in progress:
I recently learned that the Google Play app sends an intent ```com.android.vending.INSTALL_REFERRER``` to the app it just installed, containing the parameters that were part of it's install url e.g. ```https://play.google.com/store/apps/details?id=com.piwik.demo&referrer=<REFERRERPARAMTERS>```.

That started the whole coding session and while add it changed a few other things.

* ```Tracker.trackAppDownload()``` previously passed the installer packagename via QueryParams.REFERRER without modifcation, no one seems to have noticed but Piwik backend ignores non URLs.... so we now prepend ```http://```
* The library now has a receiver that listens for above mentioned intent and if received, saves it into preferences.
* Because the dev can call the track download method from within the ```Application.onCreate()``` , and the broadcast receiver causes it to initialize, the actual tracking is deferred by 3 seconds if Google Play is detected as download source.
* I removed the ```Context``` parameter from the track download methods. Previously the idea was to use this as a way to track downloads of other apps. I've come to the conclusions though that this is not very clean because we can't guarantee certain functionality for other apps (access to the apk for checksum or receiving the install referrer). It's not something we should support, a dev could make his own method for this if wanted.
* I removed the ```ExtraIdentifier.INSTALLER_PACKAGENAME``` flag. It was not a very good criteria to differentiate different downloads and this is already covered by Piwiks actual Referrer section. The default flag is now ```None```.
* The download tracking is now in a sepperate class (I don't like huge classes).
* I've added syncronization to the download tracking method that should only fire once per app version, while rare it was previously possible to cause multiple calls to execute.

I've still got to add some unit tests and do some code polish.

For now i wanted some feedback and a question.

@dotsbb What do you think about the idea of catching the Google Play intent and passing the data along with ```QueryParams.REFERRER``` field? I think the idea is cool, but i fear it's too specific and a dev should decide himself if this is something he wants to add. Maybe we should not fill the REFERRER parameter at all, by default, because we can't assume this is how most of the devs would want to use it... But then again that could be argued for tracking downloads too...

@mattab Can you tell me what exactly the ```urlref``` parameter is used for when using the HTTP API? It seems the campaign section is only filled when passing the campaign names as queryparameter or part of the main ```url```. Is ```urlref``` parsed at all? There is an extra Campaign plugin in Piwik Market, but i wasn't able to install plugins on the demo server.

testreferrerparameters

### DIFF
--- a/piwik-sdk/src/main/AndroidManifest.xml
+++ b/piwik-sdk/src/main/AndroidManifest.xml
@@ -4,4 +4,14 @@
 
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
+
+    <application>
+        <receiver
+            android:name=".InstallReferrerReceiver"
+            android:exported="true">
+            <intent-filter>
+                <action android:name="com.android.vending.INSTALL_REFERRER"/>
+            </intent-filter>
+        </receiver>
+    </application>
 </manifest>
--- a/piwik-sdk/src/main/java/org/piwik/sdk/DownloadTrackingHelper.java
+++ b/piwik-sdk/src/main/java/org/piwik/sdk/DownloadTrackingHelper.java
@@ -1,0 +1,123 @@
+package org.piwik.sdk;
+
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+import android.os.Handler;
+import android.os.Looper;
+import android.support.annotation.NonNull;
+
+import org.piwik.sdk.Tracker.ExtraIdentifier;
+import org.piwik.sdk.tools.Checksum;
+import org.piwik.sdk.tools.Logy;
+
+import java.io.File;
+
+class DownloadTrackingHelper {
+    protected static final String LOGGER_TAG = Piwik.LOGGER_PREFIX + "DownloadTrackingHelper";
+    private static final String INSTALL_SOURCE_GOOGLE_PLAY = "com.android.vending";
+    private final Tracker mTracker;
+    private final Object TRACK_ONCE_LOCK = new Object();
+    private final PackageManager mPackMan;
+    private final String mPackageName;
+    private final SharedPreferences mPreferences;
+    private final Piwik mPiwik;
+    private final Context mContext;
+
+    DownloadTrackingHelper(Tracker tracker) {
+        mTracker = tracker;
+        mPiwik = tracker.getPiwik();
+        mContext = mPiwik.getContext();
+        mPreferences = mPiwik.getSharedPreferences();
+        mPackageName = mContext.getPackageName();
+        mPackMan = mContext.getPackageManager();
+    }
+
+    void trackOnce(@NonNull ExtraIdentifier extra) {
+        try {
+            PackageInfo pkgInfo = mContext.getPackageManager().getPackageInfo(mContext.getPackageName(), 0);
+            String firedKey = "downloaded:" + pkgInfo.packageName + ":" + pkgInfo.versionCode;
+            synchronized (TRACK_ONCE_LOCK) {
+                if (!mPreferences.getBoolean(firedKey, false)) {
+                    mPreferences.edit().putBoolean(firedKey, true).apply();
+                    trackNewAppDownload(extra);
+                }
+            }
+        } catch (PackageManager.NameNotFoundException e) {
+            e.printStackTrace();
+        }
+    }
+
+    void trackNewAppDownload(@NonNull final ExtraIdentifier extra) {
+        String referringApp = mPackMan.getInstallerPackageName(mPackageName);
+        if (INSTALL_SOURCE_GOOGLE_PLAY.equals(referringApp)) {
+            Logy.d(LOGGER_TAG, "Google Play is install source, deferring tracking.");
+            // Installed by Google Play
+            // If we are called from from within Application.onCreate wait for the install REFERRER
+            new Handler(Looper.getMainLooper()).postDelayed(new Runnable() {
+                @Override
+                public void run() {
+                    trackNewAppDownloadInternal(extra);
+                }
+            }, 3000);
+        } else {
+            trackNewAppDownloadInternal(extra);
+        }
+    }
+
+    void trackNewAppDownloadInternal(@NonNull ExtraIdentifier extra) {
+        Logy.d(LOGGER_TAG, "Tracking app download...");
+        StringBuilder installationIdentifier = new StringBuilder();
+        try {
+            installationIdentifier.append("http://").append(mPackageName); // Identifies the app
+
+            PackageInfo pkgInfo = mPackMan.getPackageInfo(mPackageName, 0);
+            installationIdentifier.append(":").append(pkgInfo.versionCode);
+
+            if (extra == ExtraIdentifier.APK_CHECKSUM) {
+                ApplicationInfo appInfo = mPackMan.getApplicationInfo(mPackageName, 0);
+                if (appInfo.sourceDir != null) {
+                    String md5Identifier = null;
+                    try {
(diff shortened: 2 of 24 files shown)


dotsbb:
>What do you think about the idea of catching the Google Play intent and passing the data along with QueryParams.REFERRER field? 
@d4rken I think it's a brliant idea! Didn't knew about this `INSTALL_REFERRER` intent.
Filling REFERRER param by default produces none overhead (just 3 sec timeout). 
I would love to have such feature in Piwik SDK by default.

piwik-sdk/src/main/AndroidManifest.xml
```diff
@@ -4,4 +4,14 @@
 
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
+
+    <application>
+        <receiver
+            android:name=".InstallReferrerReceiver"
```
dotsbb: This part might be mentioned in readme.

piwik-sdk/src/main/java/org/piwik/sdk/DownloadTrackingHelper.java
```diff
@@ -0,0 +1,123 @@
...
+                }
+            }
+
+            // Usual USEFUL values of this field will be: "com.android.vending" or "com.android.browser", i.e. app packagenames.
+            // This is not guaranteed, values can also look like: app_process /system/bin com.android.commands.pm.Pm install -r /storage/sdcard0/...
+            String referringApp = mPackMan.getInstallerPackageName(mPackageName);
+            if (referringApp != null && referringApp.length() > 200)
+                referringApp = referringApp.substring(0, 200);
+
+            if (referringApp != null && referringApp.equals("com.android.vending")) {
```
dotsbb: "com.android.vending" can be replaced by `INSTALL_SOURCE_GOOGLE_PLAY` const

piwik-sdk/src/test/java/org/piwik/sdk/TrackerTest.java
```diff
@@ -248,7 +227,7 @@ public void testGetResolution() throws Exception {
         Tracker tracker = createTracker();
         TrackMe trackMe = new TrackMe();
         tracker.track(trackMe);
-        assertTrue(trackMe.build().contains("res=480x800"));
+        assertTrue(trackMe.build(), trackMe.build().contains("res=480x800"));
```
dotsbb: Please revert this line

dotsbb:
@d4rken Piwik should be proud that have contributor like You :+1:  Let me know when You are ready, I'll look at code once again.

piwik-sdk/src/test/java/org/piwik/sdk/TrackerTest.java
d4rken: Why?

piwik-sdk/src/main/AndroidManifest.xml
d4rken: So the dev is aware that Piwik registers a broadcastreceiver? Or in case the dev wants to register this receiver himself? We could rebroadcast the intent to allow registering multiple receivers for this intent.

d4rken:
I'm not satisfied with how the download tracking is deferred, i.e. it being a fixed timer, but I don't have a better idea. It works, it's low impact, just seems dirty...

piwik-sdk/src/test/java/org/piwik/sdk/TrackerTest.java
dotsbb: One unit test is failing:
```
org.piwik.sdk.TrackerTest > testGetResolution FAILED
    java.lang.AssertionError at TrackerTest.java:230
```

d4rken: but afaik not because of that change, the first parameter of assertTrue in this case is an error message, i had the same error and wanted to debug it.

d4rken: http://developer.android.com/reference/junit/framework/Assert.html#assertTrue%28java.lang.String,%20boolean%29

dotsbb: Yeah, my bad. I mixed it up with [python's one](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertTrue) which I use on a daily basis..

piwik-sdk/src/main/AndroidManifest.xml
dotsbb: nvm, I thought that devs have to declare it explicitly in theirs apps `AndroidManifest.xml`

piwik-sdk/src/test/java/org/piwik/sdk/TrackerTest.java
d4rken: The values returned are somehow 0x0, not sure why that suddenly is. I saw no changes regarding that.

dotsbb: I got this, I'll raise PR in a minutes

d4rken:
The broadcast receiver now forwards the intent instead of consuming.
Polished some code, added comments and tests.

d4rken:
@dotsbb I'm done.
You can simulate the extra referall information like this:

``` bash
adb install -i "com.android.vending" -r base.apk
adb shell am broadcast -a com.android.vending.INSTALL_REFERRER -n com.piwik.demo/org.piwik.sdk.InstallReferrerReceiver --es "referrer" "blablamyextrainfo"
```

dotsbb:
:+1: QA done, works like a charm

quba:
@d4rken `urlref` forwards the real HTTP referrer. During tracking, Piwik decides if it's a direct entry, referrer website, search engine or an entry from a social network. Campaign parameters are available in the `url` parameter and the tracking client extracts the parameters (campaign and keyword) and sends them to Piwik via `_rcn` and `_rck` parameters. Piwik is also able to extract these parameters directly from the `url` parameter, see [this line.](https://github.com/piwik/piwik/blob/e4c43d1dfd6681b75d56c9e64750352cef193597/plugins/Referrers/Columns/Base.php#L359) Then, while tracking, they are truncated and saved in separate columns in the DB.

Please note that if campaign parameter is set, it overrides standard referrer in Piwik (so the referrer for such a visit is set to `Campaign`).

d4rken:
@quba Thanks for the info. Sadly this doesn't seem apply very well to the use case of tracking an app instead of a website. Our referrer is the install source, i.e. what other app installed us and for that we potentially get some extra campaign information.

So to have Piwik automatically parseout campaign parameters we would have to append them to:
https://github.com/piwik/piwik-sdk-android/pull/83/files#diff-663d63039a040b9dfc3bebacbeb99045R126
``` java
set(QueryParams.URL_PATH, "/application/downloaded")
```
That seems a bit dirty... 

It would be cool to have the campaign field somehow populated with info from the install referrer. We could attempt to parse for common campaign parameters (e.g. Piwik defaults and Google Analytics) and then prefill ```_rcn``` and ```_rck```. Basically doing what the server does for ```url```, but client side on the INSTALL_REFERRER data when download tracking.

Or we say if a developer wants this they can extend the client themselves and add parsing or write a server plugin that parses ```urlref```.

If first dismissed the idea because I didn't like parsing for those keywords, but it doesn't seem that bad now if we elect to parse only for a few very common ones.

@dotsbb Thoughts?

quba:
What are the possible referrers for app download? Do you maybe have some examples?

d4rken:
Well on the one hand the "referrer" could be seen as the app that installed our app.
E.g. Google Play, Amazon Appstore, but also a local file explorer that installs the app from an .apk file on the sdcard could be the "install source" and thus a referrer.

In case of Google Play, we actually have a second level of referrer, meaning we not only get the app that installed us (Google Play, aka com.android.vending), but also information about what took the user to the Google Play app, e.g. some link with extra parameters.
I'm not aware of any other Android app source that forwards these kind of parameters.

Issue #89, opened 2016-03-17
Concurrency issue when setting custom variables


d4rken:
Seems to be very rare, happened only a few times out of a few 10k events.
Writing this down so I don't forget it :).
Maybe fix this and do #69 at the same time?

```
java.util.ConcurrentModificationException
        at java.util.HashMap$HashIterator.nextEntry(HashMap.java:806)
        at java.util.HashMap$EntryIterator.next(HashMap.java:843)
        at java.util.HashMap$EntryIterator.next(HashMap.java:841)
        at org.json.JSONObject.<init>(JSONObject.java:128)
        at org.piwik.sdk.CustomVariables.toString(CustomVariables.java:90)
        at org.a.a.e.a(Tracker.java:16684)
        at eu.thedarken.sdm.tools.e.a.a(PiwikHelper.java:13159)
        at eu.thedarken.sdm.tools.piwik.PiwikHelper$1.run(PiwikHelper.java:305)
        at java.lang.Thread.run(Thread.java:841)
```

dotsbb:
I would like to take care of it, if You don't mind

d4rken:
Sure :+1:, happy hacking.

d4rken:
Fixed by @dotsbb in #90

Issue #69, opened 2015-12-01
Supporting more than 5 Custom Variables


derekchan:
It is possible to configure Piwik to receive more than 5 cvars (http://piwik.org/faq/how-to/faq_17931/). 

CustomVariables.java is currently hard-coded to support 5 CVars. Should we instead make it configurable?

```    
    private static final int MAX_VARIABLES = 5;
    protected static final int MAX_LENGTH = 200;

    public CustomVariables() {
        super(MAX_VARIABLES);
    }
```

d4rken:
Sure, we could do that.
Are you currently obstructed by this limitation?

I would suggest we pass this as parameter within the Tracker constructor as it is something that doesn't and shouldn't change over the objects lifetime.
As the Tracker constructor parameters are becoming quite unwiedly we could introduce a TrackerConfig class. Each setter of this class should have javadoc that explains the option. Especially in case of the custom variables that this needs to be configured serverside too.

I'm not yet sure though how we could elegantly ensure that the CustomVariables object within the TrackMe class also adheres to this configuration.

d4rken:
Limit was removed by @dotsbb in #90

PR #90, opened 2016-03-19
Change number of custom vars and fix concurrent error

dotsbb:
* Supporting more than 5 Custom Variables #69 
* Concurrency issue when setting custom variables #89

### DIFF
--- a/piwik-sdk/src/main/java/org/piwik/sdk/CustomVariables.java
+++ b/piwik-sdk/src/main/java/org/piwik/sdk/CustomVariables.java
@@ -14,24 +14,24 @@ import org.piwik.sdk.tools.Logy;
 import java.util.Arrays;
 import java.util.HashMap;
 
-
-public class CustomVariables extends HashMap<String, JSONArray> {
-    private final Object mConcurentLock = new Object();
+/**
+ * You can track up to 5 custom variables for each user to your app,
+ * and up to 5 custom variables for each screen view.
+ * You may configure Piwik to track more custom variables: http://piwik.org/faq/how-to/faq_17931/
+ * <p/>
+ * Desired json output:
+ * {
+ * "1":["OS","iphone 5.0"],
+ * "2":["Piwik Mobile Version","1.6.2"],
+ * "3":["Locale","en::en"],
+ * "4":["Num Accounts","2"],
+ * "5":["Level","over9k"]
+ * }
+ */
+public class CustomVariables {
+    private final Object mConcurrentLock = new Object();
+    private final HashMap<String, JSONArray> mVars = new HashMap<>();
     private static final String LOGGER_TAG = Piwik.LOGGER_PREFIX + "CustomVariables";
-    /**
-     * You can track up to 5 custom variables for each user to your app,
-     * and up to 5 custom variables for each screen view.
-     * You may configure Piwik to track more custom variables: http://piwik.org/faq/how-to/faq_17931/
-     * <p/>
-     * Desired json output:
-     * {
-     * "1":["OS","iphone 5.0"],
-     * "2":["Piwik Mobile Version","1.6.2"],
-     * "3":["Locale","en::en"],
-     * "4":["Num Accounts","2"],
-     * "5":["Level","over9k"]
-     * }
-     */
     protected static final int MAX_LENGTH = 200;
 
     /**
@@ -72,25 +72,23 @@ public class CustomVariables extends HashMap<String, JSONArray> {
      * @param values packed key/value pair
      * @return super.put result or null if key is null or value length is not equals 2
      */
-    @Override
     public JSONArray put(String index, JSONArray values) {
         if (values.length() == 2 && index != null) {
-            synchronized (mConcurentLock) {
-                return super.put(index, values);
+            synchronized (mConcurrentLock) {
+                return mVars.put(index, values);
             }
         }
         Logy.d(LOGGER_TAG, "value length should be equal 2");
         return null;
     }
 
-    @Override
     public String toString() {
-        if (size() == 0) {
+        if (mVars.size() == 0) {
             return null;
         }
         JSONObject jsonObject;
-        synchronized (mConcurentLock) {
-            jsonObject = new JSONObject(this);
+        synchronized (mConcurrentLock) {
+            jsonObject = new JSONObject(mVars);
         }
         return jsonObject.toString();
     }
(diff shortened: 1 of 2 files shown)


piwik-sdk/src/main/java/org/piwik/sdk/Piwik.java
```diff
@@ -68,6 +68,16 @@ public synchronized Tracker newTracker(@NonNull String trackerUrl, int siteId) t
     }
 
     /**
+     * Configure Piwik to receive more than 5 custom variable
+     * http://piwik.org/faq/how-to/faq_17931/
+     * Changing this limit will affect all Tracker instances
+     * @param maxVariables
+     */
+    public void setMaxCustomVariables(int maxVariables){
+        CustomVariables.MAX_VARIABLES = maxVariables;
```
d4rken: I don't like the use of a static variables like this. I'm on mobile ATM so I can't browse code well, can't we pass this as configuration parameter before tracker init? What do you think?

dotsbb: @d4rken I tried to avoid static vars, but my shot wasn't successful and elegant due to [this usage of CustomVariables](https://github.com/piwik/piwik-sdk-android/blob/8e1c23d45e37f7976db5d12f72b360a313001177/piwik-sdk/src/main/java/org/piwik/sdk/TrackMe.java#L23) in `TrackMe` class, it knows nothing about `Tracker` instance.

d4rken: What about removing the limit from the custom variables object and when submitting the trackme, the trackers trims the custom variables to the allowed amount before handing it to the dispatcher?

dotsbb: I think the most elegant way will be to remove limit of custom vars completely. 
We just can assume that developers are responsible enough and read the doc strings.

piwik-sdk/src/main/java/org/piwik/sdk/CustomVariables.java
```diff
@@ -46,12 +42,14 @@ public CustomVariables() {
...
      *              will be deleted and replaced with the new custom variable stored in index 1.
+     *              You may configure Piwik to track more custom variables than 5.
+     *              Read more: http://piwik.org/faq/how-to/faq_17931/
      * @param name  of a specific Custom Variable such as "User type".
      * @param value of a specific Custom Variable such as "Customer".
      * @return super.put result if index in right range and name/value pair aren't null
      */
     public JSONArray put(int index, String name, String value) {
-        if (index > 0 && index <= MAX_VARIABLES && name != null & value != null) {
+        if (index > 0 && name != null & value != null) {
```
dotsbb: Wrong indexes will be ignored on [the server side](https://github.com/piwik/piwik/blob/3d0684cda418d51596626034b8811942ed413f38/core/Tracker/Request.php#L574-L581)

piwik-sdk/src/main/java/org/piwik/sdk/Piwik.java
d4rken: I agree, that sounds like a clean solution with some javadoc explanation.
Am 24.03.2016 00:39 schrieb "Alexey Subbotin" <notifications@github.com>:

d4rken:
What about wrapping an instance of HashMap instead of extending it? Would make it way easier to guard (just synchronization on 3 methods). Other calls like Map.clear() might still cause issues otherwise, I'm still on mobile, can't check the sources well.

We only need a few specific methods anyways and not all HashMap has to offer.

What do you think?

dotsbb:
@d4rken I've got rid of extending HashMap, as always good idea :+1:

piwik-sdk/src/main/java/org/piwik/sdk/CustomVariables.java
```diff
@@ -14,29 +14,26 @@
...
+ * "1":["OS","iphone 5.0"],
+ * "2":["Piwik Mobile Version","1.6.2"],
+ * "3":["Locale","en::en"],
+ * "4":["Num Accounts","2"],
+ * "5":["Level","over9k"]
+ * }
+ */
+public class CustomVariables {
+    private final Object mConcurrentLock = new Object();
+    private final HashMap<String, JSONArray> mVars = new HashMap<>();
```
d4rken: I think we can go with 

```
private final Map<String, JSONArray> mVars = new ConcurrentHashMap<>();
```
and then skip the manual syncronization.
According to it's documentation the map iterator is safe during multithreading as long as each thread uses it's own iterator.

piwik-sdk/src/main/java/org/piwik/sdk/CustomVariables.java
```diff
@@ -73,21 +72,25 @@ public JSONArray put(int index, String name, String value) {
      * @param values packed key/value pair
      * @return super.put result or null if key is null or value length is not equals 2
      */
-    @Override
     public JSONArray put(String index, JSONArray values) {
         if (values.length() == 2 && index != null) {
-            return super.put(index, values);
+            synchronized (mConcurrentLock) {
```
d4rken: Just a different style of coding, do you like it?
(Assuming the change to ConcurrentHashMap)

```
public JSONArray put(String index, JSONArray values) {
    if (values.length() != 2 || index == null) {
        Logy.d(LOGGER_TAG, "values.length() should be equal 2");
    	return null;
    }
    return mVars.put(index, values);
}
```

piwik-sdk/src/main/java/org/piwik/sdk/CustomVariables.java
```diff
@@ -73,21 +72,25 @@ public JSONArray put(int index, String name, String value) {
...
+            }
         }
         Logy.d(LOGGER_TAG, "value length should be equal 2");
         return null;
     }
 
-    @Override
     public String toString() {
-        if (size() == 0) {
+        if (mVars.size() == 0) {
```
d4rken: Assuming the change to ConcurrentHashMap.
```
        JSONObject json = new JSONObject(mVars);
        return json.length() > 0 ? json.toString() : null;
```

There is also a race condition issue this fixes (we have not noticed yet).
After passing the `== 0` check, the map can still be emptied before we copy it.
So then we would return an empty String instead of a `null` String.

d4rken:
Sorry took me a bit longer.
Tests look good, but I may have found a sleaker solution.
Let me know what you think regarding my comments.

Should we do another release after this is merged into dev?

On a side note, shouldn't we submit PRs via forks ;)?

piwik-sdk/src/main/java/org/piwik/sdk/CustomVariables.java
dotsbb: Good catch :+1: Thanks

dotsbb:
@d4rken all comments addressed/fixed.

> Should we do another release after this is merged into dev?

Sure! 

> On a side note, shouldn't we submit PRs via forks ;)?

Honestly I Didn't think about it before You asked, for me it was easier to maintain only remote.

d4rken:
:+1: I think this can be merged into dev now.

Then we could either do a release with that state or see if we want to include something for #92

Issue #92, opened 2016-04-07
Tracker.trackEvent should default URL to the last viewed


derek9gag:
Referring to  
https://github.com/piwik/piwik-sdk-android/blob/master/piwik-sdk/src/main/java/org/piwik/sdk/Tracker.java#L321  and  https://github.com/piwik/piwik-sdk-android/blob/master/piwik-sdk/src/main/java/org/piwik/sdk/Tracker.java#L686 

The events tracked by Tracker.trackEvent() is likely to be defaulted to "http://HOST/", since trackEvent() does not take URL as argument. This is very likely to be wrong, and this behaviour is inconsistent with Javascript and the iOS SDKs.

We should retain the URL of the last screen view, and use it for all subsequent events. This is implemented in iOS SDK (see https://github.com/piwik/piwik-sdk-ios/blob/0ddb5078f5ed4248bcac424639ff89e5b5c30aa7/PiwikTracker/PiwikTracker.m#L485)

d4rken:
I see what issue you mean.
Isn't that an issue with all methods except maybe the trackScreenView ones?
We can't use the URL of the last screen as the Tracker instance might be used by multiple threads.

I would suggest to change the method(s) to require an url argument or take a  (optional?) TrackMe and then in another PR consider adding some helper class for this.

d4rken:
@dotsbb You did not start on this yet, right?

dotsbb:
I didn't start yet. Feel free to go ahead with this one.

d4rken:
Initial issue fixed by #93 
Further SDK API changes in #94 to make this easier for devs in the future.

derek9gag:
Thank you!

Systemreiniger092016:
I'm not professionell, only user. 
Ich wrote a short Info few month ago on my notice App. May it is important for better working by  back up this App :
---------------- SD Maid ------------------
System Toybox/Burybox install.

Question: Please help me which file i need to open for back up my Android 6.0.1 Samsung Galaxy S6 Edge. 
Thank you so much withe nice greets  ... 
Mike

PR #93, opened 2016-04-17
Persist last used URL for future events

d4rken:
This addresses issue #92. 
When a new `TrackMe` reaches the tracker, we check if it has `QueryParams.URL_PATH` set.

* If it has a url parameter, we check if it is valid, possibly fix it, then store it in the default `TrackMe` object the tracker maintains (`mDefaultTrackMe`).
* If it doesn't have a url parameter, we use the parameter stored in the `Tracker`'s default `TrackMe`.

`TrackMe`'s should be threadsafe meaning no crashes or bad behavior, but we are obviously still susceptible to race conditions, e.g.:

* Thread1: Track screen1.
* Thread1: Track event1.
* Thread2: Track screen2.

`event1` could be tracked with the URL from `screen2`.
I don't see how we can fix or prevent this, this is up to the developer to do correctly.

I'm preparing another PR that refactores the whole tracking API which will also adress `trackEvent` not taking a url parameter.


dotsbb:
:+1:

PR #94, opened 2016-04-17
API rework

d4rken:
The idea of this change is to further modularize the different components and help devs make less mistakes.

I've moved all `Tracker.track*` methods except the basic `Tracker.track(TrackMe)` one into a new class named `TrackHelper` which uses the builder pattern to guide devs through creating correct `TrackMe`'s.

The entry to tracking any type of action is a static method on the TrackHelper object that takes only the **required** parameters. Optional parameters can then be chained to the object.

Examples:
* `TrackHelper.track().screen("/user/profile").with(tracker)` would be the shorthand form for most cases.
* Alternative: `TrackMe trackMe = TrackHelper.track().screen("/user/profile").build()` if we want to keep a `TrackMe` object for later.
* We can also base our new objects on old ones: `TrackHelper.track(oldTrackMe).goal(1).with(tracker)`

I think this is much more readable, because we see what arguments we are setting, we can no longer accidentally switch e.g. quantity and revenue because both are integer values in a 5 parameter long method. Required parameters are obvious as they are part of the initial method you have to call to even reach other parameters and they don't take `null` as argument.
This is also allows more flexibility if we want to extend the API because rather than creating new methods with more and more parameters, we just add a one parameter method.


d4rken:
Still wanna improve the previous "QuickTrack" options to follow the same style.
Also want to see if we can better integrate CustomDimensions into this pattern.

piwik-sdk/src/main/java/org/piwik/sdk/TrackHelper.java
```diff
@@ -0,0 +1,719 @@
...
+         */
+        public Screen variable(int index, String name, String value) {
+            mCustomVariables.put(index, name, value);
+            return this;
+        }
+
+        @Nullable
+        @Override
+        public TrackMe build() {
+            if (mPath == null) return null;
```
dotsbb: Shouldn't we return `getBaseTrackMe()` here instead of `null`?

piwik-sdk/src/main/java/org/piwik/sdk/TrackMe.java
```diff
@@ -20,7 +19,13 @@
 public class TrackMe {
     private static final int DEFAULT_QUERY_CAPACITY = 14;
     private final HashMap<String, String> mQueryParams = new HashMap<>(DEFAULT_QUERY_CAPACITY);
-    private final CustomVariables mScreenCustomVariable = new CustomVariables();
```
dotsbb: Happy to see this change!

d4rken: :grin: I'd still love to get rid of "visit custom variables" in the `Tracker`, but this is a good start.

piwik-sdk/src/main/java/org/piwik/sdk/TrackHelper.java
d4rken: In this case `path` is the **required** variable for tracking a screenview, which is why we want to "abort" if it is null.

At somepoint we have or should give an oportunity for sanity checks on the values, although we currently don't have many of those, there needs to be a place when to do so.
As we are moving the TrackMe and Tracker class to pure information carriers, the "Builder" has to do this job, either when calling the parameter functions or when calling build(), in this case i've chosen the later one, because this gives us the option to do sanity checks against parameters form different methods the dev might have called.

The idea is that build() returns a null value when it is not a valid (use senseless) request to submit for transmission. Returning `null` means that some values/*whatever* is fishy and the TrackMe that the developer tried to generate is likely invalid.

I could see two reasons for never returning a `null`-value here but both would be contrary to what we want. So for one if we return the base `TrackMe` we could further chain calls, but for readabilities sake, we want to encourage one callchain per tracking request.

It should also be noted that the "mBaseTrack" is not modified by a request. Specifically the "sub builders" such as `Screen` in this instance, don't modify the base `TrackMe` object. The idea is to allow the developer to keep a base TrackMe object as a global field with certain parameters (e.g. some plugin stuff), that he then can always uses in the builders with TrackHelper.track(baseTrackMe).

Also see extension of the `TrackMe` constructor allowing us to use `new TrackMe(oldTrackMe)`.

The other reason we might want a non `null` value here is to avoid NPE issues, because the Tracker would crash if we do `track(null)`, but either the developer does a null check or he uses the convinience method `TrackHelper......with(tracker)` which internally calls `build()` and simply does nothing if it is `null`.

Need to add some documentation on this... will do that before this would be merged, but I'm happy to get some review on this first.

dotsbb: I get it, avoiding long call-chains sounds reasonable to me to. If user wants to track some custom request there is `TrackMe` class available.

d4rken: Sorry I sometimes write too much ;).

dotsbb:
@d4rken Good work :+1: , I like the new API. I think we [should bump major](http://semver.org/#spec-item-8) version since there are some backward incompatible changes.

d4rken:
Good idea. I'll add a commit to this PR that increases the version.

After merging it into dev, we leave it there for a few days and then the current dev branch into master, as a v1.0.0 release?

I'm also getting closer to actually coding something for #16 but because it's all transparent changes, we could add something like that in a v1.1.0

PR #95, opened 2016-05-06
Added in a download tracker which you can specify the version number (for hybrid apps)

seancunningham:
Hi there,

Very quick and simple pull request, I just wanted to add an additional method which would allow you to specify your own version number for the tracking. This is specifically advantageous for hybrid mobile apps where you may update the app independent of rebuilding/resubmitting to the store. If there is anything else I need to do to get this accepted, let me know.

Cheers,


d4rken:
I think It's a good idea.

Before this can be merged there are a few things.
* Your method only uses the version parameter for the shared preference, but no matter what version argument is supplied, the actual event is still tracked with the current app version.
* Please submit your PR against the dev branch (which currently is a lot different than master)
* Create a unit test for your new function (you can check other unit tests for guidedance)
* If your commits only contain typo fixes, squash them into a single commit.
* I think the method should also accept the `ExtraIdentifier` parameter

PR #96, opened 2016-05-06
Added in version specification for app download tracking

seancunningham:
Sorry if there was a better way to update the pull request. Changes have been made as per your requests in #95. The version is now passed in via the constructor, not a new method as before, and a unit test has been made.

### DIFF
--- a/piwik-sdk/src/main/java/org/piwik/sdk/DownloadTracker.java
+++ b/piwik-sdk/src/main/java/org/piwik/sdk/DownloadTracker.java
@@ -24,6 +24,8 @@ public class DownloadTracker {
     private final String mPackageName;
     private final SharedPreferences mPreferences;
     private final Context mContext;
+    private String mVersion;
+    private PackageInfo mPkgInfo;
 
     public enum Extra {
         /**
@@ -42,6 +44,18 @@ public class DownloadTracker {
         this(tracker, new TrackMe());
     }
 
+    public void setVersion(String version) {
+        mVersion = version;
+    }
+    
+    public String getVersion(){
+        if (mVersion != null){
+            return mVersion;
+        }
+        mVersion = Integer.toString(mPkgInfo.versionCode);
+        return mVersion;
+    }
+    
     public DownloadTracker(Tracker tracker, TrackMe baseTrackMe) {
         mTracker = tracker;
         mBaseTrackMe = baseTrackMe;
@@ -50,20 +64,20 @@ public class DownloadTracker {
         mContext = piwik.getContext();
         mPackageName = mContext.getPackageName();
         mPackMan = mContext.getPackageManager();
+        try {
+            mPkgInfo = mPackMan.getPackageInfo(mPackageName, 0);
+        } catch (PackageManager.NameNotFoundException e) {
+            e.printStackTrace();
+        }
     }
 
     public void trackOnce(@NonNull Extra extra) {
-        try {
-            PackageInfo pkgInfo = mContext.getPackageManager().getPackageInfo(mContext.getPackageName(), 0);
-            String firedKey = "downloaded:" + pkgInfo.packageName + ":" + pkgInfo.versionCode;
-            synchronized (TRACK_ONCE_LOCK) {
-                if (!mPreferences.getBoolean(firedKey, false)) {
-                    mPreferences.edit().putBoolean(firedKey, true).apply();
-                    trackNewAppDownload(extra);
-                }
+        String firedKey = "downloaded:" + mPackageName + ":" + getVersion();
+        synchronized (TRACK_ONCE_LOCK) {
+            if (!mPreferences.getBoolean(firedKey, false)) {
+                mPreferences.edit().putBoolean(firedKey, true).apply();
+                trackNewAppDownload(extra);
             }
-        } catch (PackageManager.NameNotFoundException e) {
-            e.printStackTrace();
         }
     }
 
@@ -96,22 +110,16 @@ public class DownloadTracker {
 
     private void trackNewAppDownloadInternal(@NonNull Extra extra) {
         Logy.d(LOGGER_TAG, "Tracking app download...");
-        PackageInfo pkgInfo = null;
-        try {
-            pkgInfo = mPackMan.getPackageInfo(mPackageName, 0);
-        } catch (PackageManager.NameNotFoundException e) {
-            e.printStackTrace();
-        }
-        if (pkgInfo == null)
+        if (mPkgInfo == null)
             return;
 
         StringBuilder installIdentifier = new StringBuilder();
-        installIdentifier.append("http://").append(mPackageName).append(":").append(pkgInfo.versionCode);
+        installIdentifier.append("http://").append(mPackageName).append(":").append(getVersion());
 
         if (extra == Extra.APK_CHECKSUM) {
-            if (pkgInfo.applicationInfo != null && pkgInfo.applicationInfo.sourceDir != null) {
+            if (mPkgInfo.applicationInfo != null && mPkgInfo.applicationInfo.sourceDir != null) {
                 try {
-                    String md5Identifier = Checksum.getMD5Checksum(new File(pkgInfo.applicationInfo.sourceDir));
+                    String md5Identifier = Checksum.getMD5Checksum(new File(mPkgInfo.applicationInfo.sourceDir));
                     if (md5Identifier != null)
                         installIdentifier.append("/").append(md5Identifier);
                 } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


d4rken:
> Sorry if there was a better way to update the pull request.
Normally you would squash or revert your commits, then force push your changes towards your pullrequest branch. Google has more info on that.

I think setting the version via constructor is slightly restricting here. What about a method  `setVersion(@Nullable String version)` and then in the respective places of `DownloadTracker` make use of a `getVersion()` that pulls the version from the `mVersion` field if it is non null, and otherwise pulls it from the apps context. This would nice integration within the TrackHelper.

Make an extra test method for the above case.

I think then we can merge it, I might a small code polish PR, and then the current dev branch including your changes can be merged with master and we'll do a new release.

piwik-sdk/src/main/java/org/piwik/sdk/DownloadTracker.java
```diff
@@ -50,20 +57,23 @@ public DownloadTracker(Tracker tracker, TrackMe baseTrackMe) {
         mContext = piwik.getContext();
         mPackageName = mContext.getPackageName();
         mPackMan = mContext.getPackageManager();
+        try {
+            PackageInfo pkgInfo = mPackMan.getPackageInfo(mContext.getPackageName(), 0);
+            mVersion = Integer.toString(pkgInfo.versionCode);   
+            mName = pkgInfo.packageName;
+        }
+        catch (PackageManager.NameNotFoundException e) {
```
d4rken: No new line here.

piwik-sdk/src/test/java/org/piwik/sdk/DownloadTrackerTest.java
```diff
@@ -86,7 +86,22 @@ public void testTrackNewAppDownload() throws Exception {
         assertEquals("http://" + PiwikTestApplication.INSTALLER_PACKAGENAME, queryParams.get(QueryParams.REFERRER));
 
         tracker.clearLastEvent();
+        
+        downloadTracker = new DownloadTracker(tracker, "2");
+        downloadTracker.trackNewAppDownload(DownloadTracker.Extra.APK_CHECKSUM);
+        Thread.sleep(100); // APK checksum happens off thread
+        queryParams = parseEventUrl(tracker.getLastEvent());
+        checkNewAppDownload(queryParams);
```
d4rken: As extra test method please.

piwik-sdk/src/main/java/org/piwik/sdk/DownloadTracker.java
```diff
@@ -41,7 +43,25 @@
     public DownloadTracker(Tracker tracker) {
         this(tracker, new TrackMe());
     }
-
+    
+    public void setVersion(String version) {
+        mVersion = version;
+        System.out.println("Version is now "+version);
```
dotsbb: I would suggest to use `Logy.d` here instead, as we did it [here](https://github.com/piwik/piwik-sdk-android/blob/master/piwik-sdk/src/main/java/org/piwik/sdk/Tracker.java#L723).

piwik-sdk/src/main/java/org/piwik/sdk/DownloadTracker.java
```diff
@@ -50,20 +70,25 @@ public DownloadTracker(Tracker tracker, TrackMe baseTrackMe) {
         mContext = piwik.getContext();
         mPackageName = mContext.getPackageName();
         mPackMan = mContext.getPackageManager();
+        try {
+            PackageInfo pkgInfo = mPackMan.getPackageInfo(mContext.getPackageName(), 0);
+            mVersion = getVersion();
+            if (mVersion == null){
+                throw new PackageManager.NameNotFoundException("Version could not be found");
```
dotsbb: This `try/catch` block and those one above at `L56` looks very similar. It's better to refactor them and follow DRY principle

seancunningham:
Totally right, the try/catches were a bit unnecessary other than to get the package (which I don't think is necessary anymore, but followed the pattern I saw). I've refactored it to only have the try/catch in the constructor - as long as the version getter is using the same pkgInfo object, the try/catch isn't needed. Also removed that debug line -- wasn't intending to commit that and am glad it was caught.

piwik-sdk/src/main/java/org/piwik/sdk/DownloadTracker.java
```diff
@@ -50,20 +68,26 @@ public DownloadTracker(Tracker tracker, TrackMe baseTrackMe) {
         mContext = piwik.getContext();
         mPackageName = mContext.getPackageName();
         mPackMan = mContext.getPackageManager();
+        try {
+            mPkgInfo = mPackMan.getPackageInfo(mPackageName, 0);
+            mName = mPkgInfo.packageName;
```
d4rken: The packagename is already stored in `mPackageName`.

piwik-sdk/src/main/java/org/piwik/sdk/DownloadTracker.java
```diff
@@ -50,20 +68,26 @@ public DownloadTracker(Tracker tracker, TrackMe baseTrackMe) {
...
+            mVersion = getVersion();
+            if (mVersion == null){
+                throw new PackageManager.NameNotFoundException("Version could not be found");
+            }
+        }
+        catch (PackageManager.NameNotFoundException e) {
```
d4rken: No new line.

piwik-sdk/src/main/java/org/piwik/sdk/DownloadTracker.java
```diff
@@ -41,7 +44,22 @@
...
+    public void setVersion(String version) {
+        mVersion = version;
+        Logy.d(LOGGER_TAG, "Version Set");
+    }
+    
+    public String getVersion(){
+        if (mVersion != null){
+            return mVersion;
+        }
+        else if (mPkgInfo != null){
```
d4rken: No new line.

piwik-sdk/src/main/java/org/piwik/sdk/DownloadTracker.java
```diff
@@ -41,7 +44,22 @@
...
     }
-
+    
+    public void setVersion(String version) {
+        mVersion = version;
+        Logy.d(LOGGER_TAG, "Version Set");
+    }
+    
+    public String getVersion(){
+        if (mVersion != null){
```
d4rken: Missing space between the brackets.

piwik-sdk/src/main/java/org/piwik/sdk/DownloadTracker.java
```diff
@@ -50,20 +68,26 @@ public DownloadTracker(Tracker tracker, TrackMe baseTrackMe) {
...
     public void trackOnce(@NonNull Extra extra) {
-        try {
-            PackageInfo pkgInfo = mContext.getPackageManager().getPackageInfo(mContext.getPackageName(), 0);
-            String firedKey = "downloaded:" + pkgInfo.packageName + ":" + pkgInfo.versionCode;
-            synchronized (TRACK_ONCE_LOCK) {
-                if (!mPreferences.getBoolean(firedKey, false)) {
-                    mPreferences.edit().putBoolean(firedKey, true).apply();
-                    trackNewAppDownload(extra);
-                }
+        String firedKey = "downloaded:" + mName + ":" + mVersion;
```
d4rken: The idea is to use getVersion() through out the code instead of mVersion, which load either provide a manually set version or pull one from the context.

piwik-sdk/src/main/java/org/piwik/sdk/DownloadTracker.java
```diff
@@ -50,20 +68,26 @@ public DownloadTracker(Tracker tracker, TrackMe baseTrackMe) {
         mContext = piwik.getContext();
         mPackageName = mContext.getPackageName();
         mPackMan = mContext.getPackageManager();
+        try {
+            mPkgInfo = mPackMan.getPackageInfo(mPackageName, 0);
+            mName = mPkgInfo.packageName;
+            mVersion = getVersion();
+            if (mVersion == null){
```
d4rken: If `getPackageInfo` did not throw an exception then `getVersion()` is never null. A packageInfo object always has a `versionCode`. The null check and extra exception is superfluos.

piwik-sdk/src/test/java/org/piwik/sdk/DownloadTrackerTest.java
```diff
@@ -99,6 +100,39 @@ public void testTrackNewAppDownload() throws Exception {
...
+        assertEquals("http://" + PiwikTestApplication.INSTALLER_PACKAGENAME, queryParams.get(QueryParams.REFERRER));
+        
+        tracker.clearLastEvent();
+        
+        downloadTracker.setVersion(null);       
+        downloadTracker.trackNewAppDownload(DownloadTracker.Extra.APK_CHECKSUM);
+        Thread.sleep(100); // APK checksum happens off thread
+        queryParams = parseEventUrl(tracker.getLastEvent());
+        checkNewAppDownload(queryParams);
+       m = REGEX_DOWNLOADTRACK.matcher(queryParams.get(QueryParams.DOWNLOAD));
```
dotsbb: Indentation to fix

dotsbb:
:+1: Looks good to me

PR #98, opened 2016-05-16
Replacing Logy with Timber

d4rken:
Replaced selfmade Logy class and use with the more sophisticated Timber logging library.

Changed logging calls to make use of the on-demand string formatting from Timber which reduces memory impact of our library in production ( i.e. when nothing is to be logged).

If a dev already uses Timber, Piwik will now automatically take part in his logging solution without any extra effort. If a dev is not using Timber, he needs to plant a `Tree` to view the output. The example app contains an example :grin: .


d4rken:
@dotsbb Would you do a small review before I merge?

dotsbb:
Go for it

Issue #85, opened 2016-02-02
Setting connection timeout


indywidualny:
I'm trying to change connection timeout from 5 to 10 seconds. My server is not very stable at the moment. I tried `mPiwikTracker.getDispatcher().setTimeOut(10000);` in try block below but `getDispatcher()` has protected access. My first question - why? And the second: how do it properly?

I post some code from the class extending Application class to make it easier. This is how I get the tracker.

```
public synchronized Tracker getTracker() {
        if (mPiwikTracker != null)
            return mPiwikTracker;

        try {
            mPiwikTracker = Piwik.getInstance(this).newTracker("http://indywidualni.org/analytics/piwik.php", 1);
            mPiwikTracker.setUserId(Settings.Secure.getString(getContentResolver(), Settings.Secure.ANDROID_ID));
            mPiwikTracker.setDispatchInterval(-1);
        } catch (MalformedURLException e) {
            Log.w("Piwik", "url is malformed", e);
            return null;
        }

        return mPiwikTracker;
    }
```

And a piece from onCreate

```
getTracker().trackAppDownload();
getTracker().dispatch();
```

d4rken:
Hm yeah that is a problem.
The dispatcher is internal functionality, basically the ```Tracker``` is missing a setter for the timeout value.

I'll add a public method on the dev branch.

d4rken:
The dev branch now contains setters. You can build a jar from the dev branch (see docs).
The next release will likely be when #83 is done.

indywidualny:
`mPiwikTracker.setDispatchTimeout(30000);` works great. Thanks!

dotsbb:
@indywidualny Maybe it's too late but anyway:
Have you seen this plugin https://github.com/piwik/plugin-QueuedTracking? I believe it can be useful for You.

indywidualny:
Thanks. I'm just checking whether user is online. If there is network
connection I'm dispatching manually because I need only one information to
be sent. Just like I said, I'm counting app installations. The plugin is
not really necessary for me because it doesn't have to be 100% accurate. If
some data is missing that's ok. I need only most of it to know a little bit
about my users :)

pon., 8.02.2016 o 23:36 użytkownik Alexey Subbotin <notifications@github.com>
napisał:

indywidualny:
[...]

Issue #64, opened 2015-10-05
Version 1.0.0 ?


mattab:
Hi @dotsbb 

When do you think we could release version 1.0.0 of the Android SDK?

What would you like to include in this 1.0.0 release?

dotsbb:
We could release 1.0.0 version with introduced offline mode #16  and fixed deprecated classes #61 at least.
Nice thing to have is Android 6.0 support: example app and gradle settings should be updated.

d4rken:
A 1.0.0 release usually means some kind of API commitment.
So we should be happy with the way the API looks.
Are we happy?

#61 is not much work. After that we should be able to build against API 23 (if there is no unforseen conflict, sometimes testing libs throw weird errors when building against newer APIs). I might do that on sunday with a nice cup of coffee.

#16 Is more tricky. I'll post some my ideas in the ticket though.@dotsbb would love to hear your thoughts on them.

The example app still contains a url to a piwik instance which 404s. What do we do with that?
@mattab It would be nice to have a working piwik instance for users to play around with. Maybe the demo app could post specific events to the piwik demo instance ?

d4rken:
We will do 1.0.0 without offline support for now.

I'll do 2-3 PRs this week, then I would say we release 1.0.0 officially.
My TODO
* Some small code polish
* A few additional helper methods within the TrackHelper
* Replacing the current logging mechanism with Timber

Can you think of anything else we might want to do @dotsbb ?

Offline support can still be added transparently later. I've looked into offline support of a few other analytics SDKs and got some idea what we don't want to do, coupled with the points in #16, maybe I'll submit a draft and @dotsbb could give me some feedback on it. While I don't really miss an offline mode, for competitiveness it should exist.

dotsbb:
@d4rken sounds like a very good plan to me. Didn't know about Timber before. I look forward to see your draft.

d4rken:
Timber requires minApi 15 :(

d4rken:
v1.0.0 has been merged into master, tagged and I've pushed v1.0.0 to bintray.

d4rken:
@dotsbb If you have time, can you create a release entry for v1.0.0 on Github?

mattab:
Awesome work on 1.0.0 release @d4rken @dotsbb :tada: 

Congratulations and a big thank you from the Piwik community!

PR #102, opened 2016-05-25
1.0.0 release

d4rken:
Merging dev into master as first 1.0.0 release (also see #64).

* Dispatcher timeout is now configureable (#85).
* Added support for "Custom Dimensions" plugin (#70).
* Reworked download tracking with support for referrer data from Google Play.
* Removed custom var limits and fixed a concurrency error (#69 #90).
* The current url path is always set to the last tracked screenview (#92).
* New tracking API based on builder pattern.
* Added Timber as logging library.
* Updated build tools and libraries.
* Small tweaks & code polish.


d4rken:
Hm LINT check complaining about max TAG length exceeding 23 chars.
Comes from Timber, but is actually something from the base Android Logging API.
We were already exceeding the limit when using `Logy` it just didn't LINT check it.
http://stackoverflow.com/questions/28168622/the-logging-tag-can-be-at-most-23-characters/28168739

Change or ignore? @dotsbb

Maybe "PIWK" or "PWIK" or "PWK" and then not using class names but shorter words for tags would get us there.

dotsbb:
@d4rken I would keep TAG as it is now

d4rken:
K, I'll commit a lint.xml config file setting the severity from error to ignore.

Issue #105, opened 2016-05-26
Can't compile with version 1.0.0


itshnegg:
The following error occurs when I add 'compile 'org.piwik.sdk:piwik-sdk:1.0.0' to the dependencies.

Error:Failed to resolve: com.jakewharton.timber:timber:4.1.3-20160408.041006-2

d4rken:
Thanks for the report, I'm looking into it. Will likely be fixed by #106

d4rken:
Should be resolved.

PR #106, opened 2016-05-26
Changed timber version back to 4.1.2

aegis123:
When updating to piwik 1.0.0 it would force me to update timber to a snapshot version. This should fix this so that developers using the piwik sdk don't need to forcibly be upgrade to snapshot version of timber.

### DIFF
--- a/piwik-sdk/src/main/AndroidManifest.xml
+++ b/piwik-sdk/src/main/AndroidManifest.xml
@@ -1,14 +1,20 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest package="org.piwik.sdk"
-          xmlns:android="http://schemas.android.com/apk/res/android">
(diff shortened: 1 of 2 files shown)


piwik-sdk/src/main/AndroidManifest.xml
```diff
@@ -1,14 +1,20 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest package="org.piwik.sdk"
-          xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          xmlns:tools="http://schemas.android.com/tools"
+          package="org.piwik.sdk">
```
d4rken: Why did you add the package?

d4rken:
This looks good, but I'm still wondering why the snapshot version (which is specific to that one snapshot build) fails to download ( #105 ).

PR #113, opened 2016-07-22
False price formatting

alex0709:
When implementing the piwik library I have noticed that the prices were always 0 because CurrencyFormatter.priceString() formats prices with german locale to "10,00" and not to "10.00". Therefore the format was wrong according to the documentation and the price was not accepted from piwik.


d4rken:
Good catch!

PR #118, opened 2016-10-14
"Typo" in readme.md

eldk:
Hello,

I think there is a typo here.
So change return piwikTracker to return mPiwikTracker.

Thanks,

Eric


d4rken:
Indeed, thanks!

PR #132, opened 2016-10-31
Update doc : gzipped json dispatch

eldk:
Hello,

I have included DOC for gzipped JSON dispatch in README.md .

Thanks,

Eric

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ tracker.track(trackMe);
 
 The tracker by default will dispatch any pending events every 120 seconds.
 
-If 0 is used, any event will be dispatch immediatly.
+If 0 is used, any event will be dispatched immediatly.
 
 If a negative value is used the dispatch timer will never run, a manual dispatch must be used:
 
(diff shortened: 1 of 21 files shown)


README.md
```diff
@@ -201,6 +201,8 @@ tracker.track(trackMe);
 
 The tracker by default will dispatch any pending events every 120 seconds.
 
+If 0 is used, any event will be dispatch immediatly.
```
d4rken: dispatch**ed**

README.md
```diff
@@ -218,6 +220,25 @@ If a negative value is used the dispatch timer will never run, a manual dispatch
     
 ```
 
+When there is more than one event in queue, dispatch is done using a POST request with JSON data (Bulktracking).
+
+Those JSON datas may be dispatched using a Gzipped output. This may be set at app init time as this :
```
d4rken: ~~Those~~ JSON data

"as follows"

or "like this"

PR #136, opened 2016-11-12
Offline mode

d4rken:
Working on #16.


d4rken:
> We should store somekind of version code or error handling with the offline data. Future updates might break compatibility with previously stored but unsent offline data. Best case, still reuse it, in any othercase discard it.

Done.

> As we are now dealing with transmission failures, there should be some kind of system that deals with "retries".

Done

> If the piwik instance is offline, this would count as transmission failure. If we have a million app users, and our piwik instance goes offline, and then goes online again, all users transmitting all offline data would probably DDOS the piwik instance.

Variance and our default 120s dispatch interval should remedy this somewhat.

> We could gradually transmit the offline data over a time span. Would we need to store new events if there still is offline or can we send them directly? (e.g. does the server get confused if they are totally out of order)

Currently directly storing new events while we are "offline", due to queryparam `cdt` being set, the order doesn't matter.

> The local storage should be limited to a configurable size. We could add a setting that allows the dev to choose betwen storing in /files or  /cache. Within cache files may be removed automatically if space runs low, but we might loose offline data if the user chooses to clear cache.

Currently only limiting by file count and timestamps (4 hours max).

> The local storage should probably be a buffer such that if we exceed the limit, the oldest offline data is overwritten.

Done.

> An analytics framework should minimize it's impact on the app and device, so if we regain connectivity, and there is a lot of offline data, we shouldn't send it all at once, but gradually transmit the views. This will also limit the impact this has on the server.

Not sure atm how to implement throtteling nicely, where would the throtteling happen? In `Dispatcher` most likely, by limiting the amount of packets dispatched per interval, but how many packets?

d4rken:
- [ ] **Make sure items are send in order**
- [ ] Cache limit (timewise) configurable
- [ ] Default cache limit 24h

eldk:
Hello,

Should you create a dev branch with this PR or merge it to dev?

I have created a branch with it on my repo (https://github.com/eldk/piwik-sdk-android/tree/dev+d4rken-pr-offlinemod), and test it on some of my app users (since yesterday).

As I want to test a wifi only mode (app wait for wifi only connection to push datas or loose it if wifi is not available < 24 hours), I think it will be more convenient.

Thanks,

Eric

d4rken:
The code isn't very good yet and we have no tests either.
I'll commit a few more changes, but it will still need a review.

d4rken:
Fixed/Implemented the outstanding points.
Also introduced the `Event` class with its own timestamp as the timestamp (for caching) should not be reset by enqueuing/dequeueing events.

We have some tests now and I think the code can work, but a few more eyes should definitely check it.

I'm wondering if caching should be enabled by default.
Current limits are `24h` or `4MB`.

d4rken:
Size based test needs to be fixed.

Eventcontainer need to be named after the last item in a cached list, such that the pre-read container check does not discard container that may still contain valid events.

d4rken:
Enabling this by default is a pretty huge change so this needs more reviews. Looking for feedback on my unit-tests, are all edge cases covered? and if my save/load code in `EventDiskCache` is sane... 😁

d4rken:
@eldk Maybe try this new branch again, it's a lot different from the one you tried previously.

eldk:
@d4rken , I will get your https://github.com/d4rken/piwik-sdk-android/tree/pr-offlinemode branch today, thanks

eldk:
Hello,
@d4rken, 
Add support for "Dispatch when connected to WIFI only" when dispatches events.
May be set to true in initPiwik() with getTracker().setDispatchWIFIOnly(true).
https://github.com/eldk/piwik-sdk-android/commits/myDev2%2Bd4rken-pr-offlinemod-wifionly

I will do tests and one enhancement  : App user should choose to send data at anytime.
Then add doc in the next days.

Thanks,

Eric

d4rken:
>@d4rken Excellent 🥇 I did a code review carefully - it all looks gorgeous.

Awesome, thanks!

> However I'll suggest to hold before merging yet. I'm going to build and .apk and test it in on my phones manually.

Good idea, will wait.

> One more thing: can You clean commented parts before merging?

What do you mean?

dotsbb:
@d4rken by commented code I mean this: https://github.com/piwik/piwik-sdk-android/pull/136/files#diff-e1da83224c84a651c268eaf45dea8a96R135

d4rken:
Ah right, sure, but now I wonder why i commented it out...

d4rken:
@eldk Did you do anymore testing? Did anything come up?

eldk:
@d4rken ,
>Did you do anymore testing?
Yes, on real devices from API 10 to API 25, It's OK, works as expected.
- if connected or connected delay <24h00, send datas
- loose last datas if delay to connexion is > 24h00 or datas size is > 4 mo.

I will work on send data when connected to wifi only mode PR until monday, as I have to release a new package for my app.
- Dev want to force WIFI only to send datas (https://github.com/eldk/piwik-sdk-android/commit/ed817dbc6a55873c72fe078fa8252b74336177ab) - user have no choice.
- Dev allow user to choose delivery mode  (https://github.com/eldk/piwik-sdk-android/commit/f05654ea558858bc607b082cfe7a9523a09005c9)
 -- wifi only
-- no matter (so wifi or mobile datas)

Thanks,

Eric

d4rken:
I've got some code ready for WIFI only, just finishing unit tests, I'll make a PR in a few mins.
Would love your feedback on it.

eldk:
OK

mattab:
I'm late to this, but well done guys for implementing Offline tracking, that's a awesome feature to have!

PR #145, opened 2017-02-02
Add options for different dispatching modes.

d4rken:
Currently supported: ALWAYS and WIFI_ONLY.
Also some light refactoring to improve testing.

### DIFF
--- a/piwik-sdk/src/main/java/org/piwik/sdk/Tracker.java
+++ b/piwik-sdk/src/main/java/org/piwik/sdk/Tracker.java
@@ -288,7 +288,8 @@ public class Tracker {
     }
 
     /**
-     * Whether the dispatch will only dispatch items on WIFI.
+     * The current dispatch behavior.
+     * @see DispatchMode
      */
     public DispatchMode getDispatchMode() {
         String raw = getSharedPreferences().getString(PREF_KEY_DISPATCHER_MODE, null);
@@ -302,7 +303,6 @@ public class Tracker {
 
     /**
      * Sets the dispatch mode.
-     *
      * @see DispatchMode
      */
     public void setDispatchMode(DispatchMode mode) {
--- a/piwik-sdk/src/main/java/org/piwik/sdk/Tracker.java
+++ b/piwik-sdk/src/main/java/org/piwik/sdk/Tracker.java
@@ -11,9 +11,11 @@ import android.content.SharedPreferences;
 import android.support.annotation.NonNull;
 import android.support.annotation.VisibleForTesting;
 
+import org.piwik.sdk.dispatcher.DispatchMode;
 import org.piwik.sdk.dispatcher.Dispatcher;
 import org.piwik.sdk.dispatcher.EventCache;
 import org.piwik.sdk.dispatcher.EventDiskCache;
+import org.piwik.sdk.tools.Connectivity;
 import org.piwik.sdk.tools.DeviceHelper;
 
 import java.net.MalformedURLException;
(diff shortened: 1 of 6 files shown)


d4rken:
@eldk I'm wondering whether "forceDispatch()" should behave like "ALWAYS", i.e. overriding any dispatchmode.

dotsbb:
@d4rken ping me whenever it's ready for code review

d4rken:
@dotsbb basically now, except for my question whether `forceDispatch` should override the behavior, what are your thoughts on this? Or should this be an extra "mode", WIFI_ONLY_BUT_ALLOW_FORCEDISPATCH? 

Hm where is `forceDispatch` most used? In low memory events and when as use-case also when closing activities... so forceDispatch would be called quite often... So forceDispatch should NOT override the DispatchMode behavior.

On a side note, I'll do another PR after this refactoring some because I noticed we can't easily inject a mock `Dispatcher` into the tracker for testing... Need to change a few constructors and decouple a bit more. But also because the coupeling between Piwik<>Tracker<>Dispatcher needs to be cleaner...

d4rken:
Ok `forceDispatch` should not override the behavior, lets keep it as it is in the PR. @dotsbb ping 😉

eldk:
>So forceDispatch should NOT override the DispatchMode behavior.
It's OK

eldk:
Hello,

Is there a way to flush cache ?

User is optin and datas are waiting to be dispatched.
Then switch to optout.
Remain cached data are still dispatched or waits for accurate connexion mode.

I think that when user optout, cache must be erased, so no more datas are send ?

So, when switch to optout, erase cache.

What do you think ?

I'm looking at it.

Eric

d4rken:
Opt-out affects whether a `track(...)` operation goes through. I don't think changing the opt-out setting should have a retroactive effect on what has already been _tracked_ and is just waiting to be dispatched.

When `track(...)` has been called, the user/dev should assume that the event will be tracked, only when and how can be affected after this.

eldk:
Hello,

Yes, it have. I just produce it when testing.

d4rken:
I think you didn't understand my post.

The way it is now is better in my opinion.
Opt-out should only affect `Tracker.track(...)`.
From the POV of users/dev, the tracking happens when they call `Tracker.track(...)`.

```
track1 -> send
track2 -> send
opt-out
track3 -> not send
track4 -> not send
opt-in
track5 -> send
track6 - send
```

makes more sense than:

```
track1 -> not send
track2 -> not send
opt-out
track3 -> not send
track4 -> not send
opt-in
track5 -> send
track6 - send
```

Especially because whether track1 and track2 is transmitted, depends on the users dispatch interval setting. In your idea (2nd example), if the dispatch interval would be set to 1ms, then track1 and track2 would have been send before the users managed to opt-out.

eldk:
Hello,

I think optout should not send or keep datas after optout.

optin
track1->stored
track1->dispatched
track2->stored
track3->stored
optout
track2&3->dispatched
track4->no store, no dispatched
optin
track5->stored
track6->stored
then dispatched
..

d4rken:
Please check my previous example again and try to understand why we should only consider opt-out in relation to `Tracker.track(...)` and not let the `Dispatcher` be concerned with the opt-out settings.

Make a new issue ticket about this if you want to discuss this further.

eldk:
Hello,
I understand what you mean, but as user and dev, even if datas have been collected before I optout, I don't want to see some datas send after that.
Thanks,
Eric

d4rken:
I see your point, but making the `Dispatcher` aware of opt-out is still not the right move in my opinion, there are other solutions for this tough. In any case, make a new ticket if you want to discuss this 😉 .

PR #147, opened 2017-02-04
Dont dispatch after optout

eldk:
Hello,

When user switch to optout, remaining events that are waiting to be dispatched need to be loose.

This is a proposal.

Maybe choice should be given to dev.

From my point of view, and the legal one (in EU), when a user optout, remaining cached datas must not be sent, even if they have been collected before the change.

Thanks,

Eric

### DIFF
--- a/piwik-sdk/src/main/java/org/piwik/sdk/dispatcher/Dispatcher.java
+++ b/piwik-sdk/src/main/java/org/piwik/sdk/dispatcher/Dispatcher.java
@@ -230,7 +230,7 @@ public class Dispatcher {
 
         if (!mDryRunOutput.isEmpty()) mDryRunOutput.clear();
 
-        if (mTracker.getPiwik().isOptOut() && !mTracker.getDispatchAfterOptout() && packet.getPostData() != null){
+        if (mTracker.getPiwik().isOptOut() && mDispatchAfterOptout && packet.getPostData() != null){
             // don't send remaining datas after user is optout. Loose them.
             Timber.tag(LOGGER_TAG).d("Optout, free cache, false dispatch");
             return true;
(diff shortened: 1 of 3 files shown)


d4rken:
I've thought about this again and while I disagree with this being a "law" issue, I agree that the behavior would be confusing to users. Your favored behavior goes well with Piwik being more privacy oriented. 

I've created #149 to stay open until your idea is implemented.

Wait for #148 though.
Either I add it in #148, or you submit a new PR when it is merged.

piwik-sdk/src/main/java/org/piwik/sdk/dispatcher/Dispatcher.java
```diff
@@ -215,6 +230,12 @@ public boolean dispatch(@NonNull Packet packet) throws IOException {
 
         if (!mDryRunOutput.isEmpty()) mDryRunOutput.clear();
 
+        if (mTracker.getPiwik().isOptOut() && !mDispatchAfterOptout && packet.getPostData() != null){
```
d4rken: This would work, but i think it's better in the order that is described in #149

piwik-sdk/src/test/java/org/piwik/sdk/dispatcher/DispatcherTest.java
```diff
@@ -128,6 +133,59 @@ public void testDispatch_gzip() throws Exception {
...
+    }
+
+    @Test
+    public void testAfterOptoutLooseRemainingDatas() throws Exception {
+        when(tracker.isDryRun()).thenReturn(false);
+        when(tracker.getPiwik()).thenReturn(piwik);
+        when(tracker.getPiwik().isOptOut()).thenReturn(true);
+        dispatcher.setDispatchAfterOptout(false);
+
+        Packet packet = mock(Packet.class);
```
d4rken: You mock a packet here, but it's just a new mock packet, it won't have any influence on the internal `Packet` objects created by `Dispatcher`.

We need better architecture to test this. I'm working on allowing better testing in #148

PR #153, opened 2017-02-11
If the system supplies no nice user-agent, build one.

### LINKED ISSUE #152: Improve device detection, manually build user agent.
19% of my traffic is listed as "Unknown device", while some of it is just failing detection (trying to improve that with https://github.com/piwik/device-detector/issues/5576), most of it because of bad data:

```
"Apache-HttpClient/UNAVAILABLE (java 1.4)"
```

While most others look like:

```
"Dalvik/2.1.0 (Linux; U; Android 6.0.1; SM-G935F Build/MMB29K)"
```

Currently we just read the system property `http.agent`, but we could create this user agent.
We can look up the dalvik version (even though it's really the ART version), we know it's `Linux; U;` we can also look up the Android version, as well as the model name and build name from https://developer.android.com/reference/android/os/Build.html

Thoughts?

d4rken:
I collected ~750k unknown devices, out of those 10% were unknown due to supplying no good user agent.
This PR adds user agent generation if none is available.
The agent is analog to most of the agents I've seen during collection and should hopefully work nicely with piwiks device-detector module.

The other 90% will require regex improvements.
A few popular ones are already covered in https://github.com/piwik/device-detector/pull/5586
The rest will require lots of work 😦, see https://github.com/piwik/device-detector/issues/5585

Closes #152.

### DIFF
--- a/piwik-sdk/src/main/java/org/piwik/sdk/tools/DeviceHelper.java
+++ b/piwik-sdk/src/main/java/org/piwik/sdk/tools/DeviceHelper.java
@@ -26,8 +26,14 @@ import timber.log.Timber;
 public class DeviceHelper {
     private static final String LOGGER_TAG = Piwik.LOGGER_PREFIX + "DeviceHelper";
     private final Context mContext;
+    private final PropertySource mPropertySource;
+    private final BuildInfo mBuildInfo;
 
-    public DeviceHelper(Context context) {mContext = context;}
+    public DeviceHelper(Context context, PropertySource propertySource, BuildInfo buildInfo) {
+        mContext = context;
+        mPropertySource = propertySource;
+        mBuildInfo = buildInfo;
+    }
 
     /**
      * Returns user language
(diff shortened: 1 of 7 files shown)


piwik-sdk/src/main/java/org/piwik/sdk/tools/DeviceHelper.java
```diff
@@ -44,7 +50,19 @@ public String getUserLanguage() {
...
-        return System.getProperty("http.agent");
+        String httpAgent = mPropertySource.getHttpAgent();
+        if (httpAgent == null || httpAgent.startsWith("Apache-HttpClient/UNAVAILABLE (java")) {
+            String dalvik = mPropertySource.getJVMVersion();
+            if (dalvik == null) dalvik = "0.0.0";
+            String android = mBuildInfo.getRelease();
+            String model = mBuildInfo.getModel();
+            String build = mBuildInfo.getBuildId();
+            httpAgent = String.format(Locale.US,
+                    "Dalvik/%s (Linux; U; Android %s; %s Build/%s)",
```
d4rken: @sgiehl Thoughts on the format? Seems to be the most common Android http agent format.

sgiehl: Guess that should work fine

PR #158, opened 2017-03-24
Allow custom identifiers when tracking downloads.

d4rken:
Make it possible to create custom extra identifiers for download tracking.

Instead of passing enums we pass objects that implement the `Extra` interface.

Now we can also move our `Extra.Checksum` code out of the core download tracker code into an `ApkChecksum extends Extra`, much cleaner 😄 .

My idea was that I wanted to use the checksum as extra identifier, but also reuse that checksum somewhere else so it's not calculated twice. With this change it is now possible to do that.

### DIFF
--- a/piwik-sdk/src/main/java/org/piwik/sdk/extra/DownloadTracker.java
+++ b/piwik-sdk/src/main/java/org/piwik/sdk/extra/DownloadTracker.java
@@ -1,6 +1,7 @@
 package org.piwik.sdk.extra;
 
 
+import android.content.Context;
 import android.content.SharedPreferences;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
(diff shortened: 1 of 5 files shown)


piwik-sdk/src/main/java/org/piwik/sdk/extra/DownloadTracker.java
```diff
@@ -23,38 +24,118 @@
...
+            Timber.tag(LOGGER_TAG).e(e);
+            throw new RuntimeException(e);
         }
     }
 
+    public DownloadTracker(Tracker tracker, @NonNull PackageInfo packageInfo) {
+        mTracker = tracker;
+        mContext = tracker.getPiwik().getContext();
+        mPreferences = tracker.getPreferences();
+        mPackMan = tracker.getPiwik().getContext().getPackageManager();
```
dotsbb: `tracker.getPiwik().getContext()` could be replaced with simply `mContext` here

d4rken: Future maintenance 😁

Issue #163, opened 2017-05-15
TrackerHelp.BaseEvent build method returns null


aegis123:
When using the `TrackerHelper` to build Events the build method is annotated with `@Nullable`. This is a bit weird since you are using a the builder pattern to create null objects which should not be possible.

The proper way would be to throw an exception at either the setter or in the build method to notify the developer he is creating build an object with an inconsistent state. Also Proper usages of `@NonNull` and `@Nullable` annotations in the arguments can help the developer with supplying the proper value to the method.

d4rken:
Something like #164 ?

The idea behind requiring null checks was that if the developer null checks these events, there would be less of a chance that unexpected tracking input crashes the app.

I agree with you though that this is counter intuitive to the builder pattern.

I'm not sure yet whether checking within the setters/constructors or in the `build()` call is the right move. Within `build()` would give the advantage that later changes to the methods wouldn't require extra checks because it still all goes through `build()`. Though one could also argue that it's better to fail early...

`Nullable` and `NonNull` annotation can still be improved in code, feel free to send a PR. Generally we should assume that a method argument/returnval is `NonNull` by default and otherwise annotate it (annotating everything with `NonNull` is a lot of spammy boilerplate).

d4rken:
Changed in #164.

PR #164, opened 2017-05-15
TrackHelper using exceptions instead of returning null.

d4rken:
PR regarding discussion in #163 

> Builder pattern should not require a null-check by default, use exception where necessary.

### DIFF
--- a/piwik-sdk/build.gradle
+++ b/piwik-sdk/build.gradle
@@ -1,6 +1,6 @@
 apply plugin: 'com.android.library'
 
-def versionMajor = 2
+def versionMajor = 3
 def versionMinor = 0
 def versionPatch = 0
 def myVersionCode = versionMajor * 10000 + versionMinor * 100 + versionPatch
--- a/piwik-sdk/src/main/java/org/piwik/sdk/extra/TrackHelper.java
+++ b/piwik-sdk/src/main/java/org/piwik/sdk/extra/TrackHelper.java
@@ -54,6 +54,9 @@ public class TrackHelper {
             return mBaseBuilder.mBaseTrackMe;
         }
 
+        /**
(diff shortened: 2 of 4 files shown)


d4rken:
I wouldn't mind. Either `3.0.0` or `2.1.0` would be fine with me.

Is it a breaking change though? The nullchecks become superfluos, but nothing previously guaranteed that `TrackHelper` methods don't throw exceptions.

d4rken:
Need to move the exception checking to the build command and then check if we can just try catch exceptions within the trackhelper to achieve something similar to the previous behavior (current one in 2.0) where bad data doesn't necessarily lead to a crash.

piwik-sdk/src/main/java/org/piwik/sdk/extra/TrackHelper.java
```diff
@@ -55,16 +54,39 @@ TrackMe getBaseTrackMe() {
...
+            return safelyWith(piwikApplication.getTracker());
+        }
+
+        /**
+         * {@link #build()} can throw an exception on illegal arguments.
+         * This can be used to avoid crashes when using dynamic {@link TrackMe} arguments.
+         *
+         * @return false if an error occured, true if the TrackMe has been submitted to be dispatched.
+         */
+        public boolean safelyWith(Tracker tracker) {
```
d4rken: @dotsbb What do you think about this?

dotsbb: Good one!

piwik-sdk/build.gradle
```diff
@@ -1,6 +1,6 @@
 apply plugin: 'com.android.library'
 
-def versionMajor = 2
```
dotsbb: :+1:

PR #168, opened 2017-06-15
Update project dependencies

friederbluemle:
This updates all project dependencies to the latest versions (as of Jun 15, 2017). It also contains a few clean ups in separate commits.

<sub><sup>Please NO GitHub rebase/squash merges. If you'd like me to combine some or all of the commits, please let me know, and I will update my branch. Thank you.</sup></sub>


d4rken:
Hm why doesn't travis kick off...

d4rken:
I think it doesn't kick off again when switching base branches, can you resubmit against `dev`?
Looks good otherwise.

friederbluemle:
Thanks for the review @d4rken Sure thing, I rebased all of my commits on top of the latest `dev` branch and force pushed my branch.
Just curious about the branching strategy in this repo: I noticed Git tags are already used to track releases (which is a good thing!). In that case, is there a need to maintain a separate `dev` branch next to the standard `master` branch? Besides the additional management effort, the required merges between the two branches create a non-linear and slightly confusing history, as seen below.
And also another minor thing I noticed related to that: The tag `v2.0` should probably be named `v2.0.0`, as that's the full semantic versioning name (which was also published to bintray).

```
$ git log -11 --graph --oneline
*   c081b54 (HEAD -> master, origin/master, origin/HEAD) Merge pull request #160 from piwik/dev
|\
| *   7d7eb83 (origin/dev, dev) Merge branch 'master' into dev
| |\
| |/
|/|
* | ac53363 Typo in README.md
* |   138f8de (tag: v1.0.2) Merge pull request #115 from piwik/dev
|\ \
| | *   8441175 Merge pull request #159 from piwik/2.0
| | |\
| | | * e1623f5 (tag: v2.0) Updated build tooling & support lib
| | | *   f84fc07 Merge pull request #158 from d4rken/pr-custom-download-identifiers
| | | |\
| | | | * ac2f630 Support tracking 3rd party apps.
| | | | * 401f0a3 Allow custom identifiers when tracking downloads.
| | | |/
| | | *   53fb110 Merge pull request #157 from d4rken/pr-trackerconfig
| | | |\
| | | | * 8092419 Introducing `TrackerConfig`
| | | |/
```

d4rken:
> In that case, is there a need to maintain a separate dev branch next to the standard master branch? Besides the additional management effort, the required merges between the two branches create a non-linear and slightly confusing history, as seen below.

`master` is what's stable (usually equal to the latest production release), while `dev` contains the bleeding edge version which may be unstable.  If we just have `master` and designate what's "stable" only via tags, then things get confusing if we need to hotfix something in a current production release and the "next update" isn't quite ready yet.. Also need a branch to merge all into PRs, to see how they play along with each other, which may be buggy so we wouldn't want that on master.

Could probably replace `dev` with a `update-v2.0.1` each update, but then we would have to decide on the next version name prematurely 😁 .

> And also another minor thing I noticed related to that: The tag v2.0 should probably be named v2.0.0, as that's the full semantic versioning name (which was also published to bintray).

Good catch, fixed that.

PR #171, opened 2017-08-20
Update Gradle wrapper to 4.1

friederbluemle:


dotsbb:
@friederbluemle 
Could you update binary wrapper to the one from official gradle repo: https://github.com/gradle/gradle/blob/v4.1.0/gradle/wrapper/gradle-wrapper.jar

with following checksum
```
sha256sum gradle-wrapper.jar 
2fe666278c4c937bef200572781334e4513237f74ad4c93f29247a3273f8b12b
```

friederbluemle:
@dotsbb Thanks for the reply. The `gradle-wrapper.jar` you linked is not the final version `4.1`. It is the wrapper used by the Gradle project itself. They used to build `4.1`, and it happens to be `4.1-rc-1`, as you can see by looking at the associated .properties file:
https://github.com/gradle/gradle/blob/v4.1.0/gradle/wrapper/gradle-wrapper.jar
The `gradle-wrapper.jar` file I included in my PR was *auto-generated* by the `wrapper` task. There is a timestamp file inside of this jar, which is why the checksum *changes every time* it is generated. You can easily reproduce this locally by repeatedly running:

    ./gradlew wrapper --gradle-version 4.1 --distribution-type all

d4rken:
I think the issue is that someone could submit a malicious binary (unknowingly or intentionally)?

dotsbb:
@d4rken yes. but in the same time I think about using something like https://github.com/int128/gradleupdate - Automatic Gradle Updater

d4rken:
> @d4rken yes. but in the same time I think about using something like https://github.com/int128/gradleupdate - Automatic Gradle Updater

Google Cloud dependency doesn't sound too nice. 

Anyways, we can merge this, while one should always be suspicious of binary files, I trust @friederbluemle here.

Issue #169, opened 2017-06-27
Cannot provide custom dispatcher


dipendrapkrl:
First of all, thank you very much for this open source gem. 
So, I was trying to use my own dispatcher and then I realized that I cannot do it. 
Looking at the source of `Piwik` class; it is singleton class which means that I cannot subclass it(it has private constructor). 
But the method `getDispatcherFactory()` in the same class is `protected` probably meant to be exposed to the subclass. 
Doesn't these two statement contradict each other? I think if the Piwik class is not singleton, it would make the sdk open to customization.
Any help would highly be appreciated.

d4rken:
There was no intention to allow custom dispatchers. The `DispatcherFactory` was introduced to improve unit testing.

It wouldn't be out of this projects scope to allow custom dispatchers though. Probably not via subclassing `Piwik` but rather by providing a default dispatcher and a `setDispatcherFactory`. Need to also decide on an API for that and change `Dispatcher` to an interface.

What's your use case for this feature?

dipendrapkrl:
Yes. the Setter for DispatcherFactory would be nice. I want the traffic to go through my own okhttp client which is the reason I am looking for custom DispatcherFactory.

d4rken:
What advantage does using your own okhttp client have?

dipendrapkrl:
Since I am using okhttp to route all the traffic that generates from my app, I would prefer this library also allow me to do so with some work but without a fork. The requirement for me is that the server that I am sending metrics expects auth tokens in the header of the request. I dont see how I can do so without my own dispatcher.

d4rken:
That's a good use case. I'll look into making this possible.

dipendrapkrl:
Thank you very much.

DavidBcn:
it is important to distinguish metrics requests with low priority from other important request that the application needs to work. Having custom dispatch makes sense to allow any application to extend in any way.

d4rken:
Please see #172, does that suffice for your use case @dipendrapkrl ?

dipendrapkrl:
Thanks @d4rken. This totally helps providing custom dispatcher for my use case. Also since Piwik SDK auto appends to the provided URL(appends piwik.php), it would be great if `Packet.getTargetURL()` would be `public` instead of `protected`. This is required because we want to know where to send the data by the custom dispatcher. Right now I am using reflection to set custom dispatcher and to get the target url. 
Thanks for the great work.

d4rken:
[...]

PR #172, opened 2017-08-24
Make it possible to supply custom dispatcher implementations

d4rken:
For #169

### DIFF
--- a/piwik-sdk/build.gradle
+++ b/piwik-sdk/build.gradle
@@ -28,6 +28,8 @@ dependencies {
     testCompile 'org.hamcrest:hamcrest-library:1.3'
     testCompile 'org.hamcrest:hamcrest-integration:1.3'
 
+    testCompile 'com.squareup.okhttp3:mockwebserver:3.8.1'
+
     // Mocktio
     testCompile 'org.mockito:mockito-core:2.2.9'
     testCompile 'org.json:json:20140107'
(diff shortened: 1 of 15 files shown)


d4rken:
Travis failed, but it's unrealated to this. Fix for another PR, #173

dotsbb:
@d4rken Strange thing. I've tried to rebuild this branch on TravicCI few times and this test failed every time.

d4rken:
I don't see how my changes would impact this test though... ?

What are your thoughts on the changes anyways.

dotsbb:
@d4rken I like the changes. I think that there is something wrong in with following logic:
image

d4rken:
Those changes are not part of this PR though?

I made no functional changes it's just refactoring to introduce a dispatcher interface.

dotsbb:
@d4rken right, I compared to with `2.0` branch instead of `master`. So it makes it even more mystic.

d4rken:
@dotsbb It was a race conditions due to the timings used to wait for events to happen.

I've extracted the code that actually sends data into an extra class with an interface that defines it. Now it's possible to use a custom implementation just to send stuff, but still use the default dispatcher and caching mechanism.

So instead of creating a custom `Dispatcher`, devs can just create a custom `PacketSender`

Also added a few unit tests that are now possible and fixed the flaky test with it.

Now I'm not sure whether I overcomplicated it...

d4rken:
@dipendrapkrl thoughts?

dipendrapkrl:
Looks awesome. Thank you @d4rken

Issue #174, opened 2017-08-31
Improve dispatch behavior when firewalled


d4rken:
e.g. Nexus9 with AFWall+

```
D/SDM:PIWIK:EventCache: Switched state to ONLINE, uncached 0 events from disk.
D/SDM:PIWIK:Dispatcher: Drained 1 events.
D/SDM:PIWIK:Dispatcher: java.net.ConnectException: Connection refused
                            at java.net.PlainSocketImpl.socketConnect(Native Method)
                            at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:334)
                            at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:196)
                            at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:178)
                            at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:356)
                            at java.net.Socket.connect(Socket.java:605)
                            at com.android.okhttp.internal.Platform.connectSocket(Platform.java:113)
                            at com.android.okhttp.Connection.connectSocket(Connection.java:196)
                            at com.android.okhttp.Connection.connect(Connection.java:172)
                            at com.android.okhttp.Connection.connectAndSetOwner(Connection.java:367)
                            at com.android.okhttp.OkHttpClient$1.connectAndSetOwner(OkHttpClient.java:130)
                            at com.android.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:329)
                            at com.android.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:246)
                            at com.android.okhttp.internal.huc.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:457)
                            at com.android.okhttp.internal.huc.HttpURLConnectionImpl.getResponse(HttpURLConnectionImpl.java:405)
                            at com.android.okhttp.internal.huc.HttpURLConnectionImpl.getResponseCode(HttpURLConnectionImpl.java:521)
                            at com.android.okhttp.internal.huc.DelegatingHttpsURLConnection.getResponseCode(DelegatingHttpsURLConnection.java:105)
                            at com.android.okhttp.internal.huc.HttpsURLConnectionImpl.getResponseCode(HttpsURLConnectionImpl.java)
                            at org.piwik.sdk.dispatcher.Dispatcher.dispatch(Dispatcher.java:256)
[...]

PR #178, opened 2017-11-22
Java8, Lints and #174

d4rken:
* I've added a backoff mechanism to address #174 . If the transmission fails we start increasing the dispatch interval from the second failure on up to 5 times. I added unit tests for this and have rewritten a couple of other unit test in `DefaultDispatcherTest`. More mocking, less dependency on other classes.
```
Dispatchintervall=100ms
Send -> Fail -> Sleep 100ms -> RetryCnt=1
Send -> Fail -> Sleep 100ms -> RetryCnt=2
Send -> Fail -> Sleep 100ms + 2*100ms -> RetryCnt=3
Send -> Fail -> Sleep 100ms + 3*100ms -> RetryCnt=4
Send -> Fail -> Sleep 100ms + 4*100ms -> RetryCnt=5
Send -> Fail -> Sleep 100ms + 5*100ms -> RetryCnt=6
Send -> Fail -> Sleep 100ms + 5*100ms -> RetryCnt=7
etc.
```
* The current dev branch aims for a `3.0.0` release thus I think it's okay to make some breaking changes. We raised the minSDK in #177 too. As Android Studio 3.0+ introduces official use of Java8 features I think we can make use of them in this lib too. I've thus added
```
    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
```
* Based on this change I went through the code and fixed LINTs.
* I refactored the unit tests to move the "Helper" classes out of the main package
* I made every class extend a `BaseTest` class which plants a `JUnitTree` so unit tests log into console too.

### DIFF
--- a/piwik-sdk/src/test/java/org/piwik/sdk/LegacySettingsPorterTest.java
+++ b/piwik-sdk/src/test/java/org/piwik/sdk/LegacySettingsPorterTest.java
@@ -6,16 +6,17 @@ import android.content.SharedPreferences;
 
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
 import org.mockito.Mock;
-import org.mockito.MockitoAnnotations;
-import org.mockito.invocation.InvocationOnMock;
+import org.mockito.junit.MockitoJUnitRunner;
 import org.mockito.stubbing.Answer;
 
 import java.util.HashMap;
 import java.util.Map;
 
+import testhelpers.BaseTest;
+
 import static org.mockito.ArgumentMatchers.anyBoolean;
-import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.anyLong;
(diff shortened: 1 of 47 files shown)


piwik-sdk/src/test/java/org/piwik/sdk/LegacySettingsPorterTest.java
```diff
@@ -180,12 +175,7 @@ public void testDownloadMapping_exists() {
         String key3 = "testkey2";
         map.put(key3, 123465);
 
-        when(mPiwikPrefs.getAll()).thenAnswer(new Answer<Map<String, ?>>() {
-            @Override
-            public Map<String, ?> answer(InvocationOnMock invocation) throws Throwable {
-                return map;
-            }
-        });
+        when(mPiwikPrefs.getAll()).thenAnswer((Answer<Map<String, ?>>) invocation -> map);
```
dotsbb: :+1:

Issue #204, opened 2018-04-18
Why Timber is included in release?


elevenetc:
Getting an exception:
```
java.lang.NoClassDefFoundError: Failed resolution of: Ltimber/log/Timber;
```
Why `Timber` is in released version of library? It's assumed that every developer uses `Timber`? Using `3.0.1` version.

d4rken:
I'll try to reproduce this. Possibly related to #184

d4rken:
Hmmm

https://github.com/bintray/gradle-bintray-plugin/issues/197

Issue #184, opened 2017-12-26
there is no logs for release apk


fetchingly:
the piwik service has no data  for the release APK,but debug APK has

d4rken:
I don't know what you mean by that.

d4rken:
We use `Timber` for logging, see:
https://github.com/JakeWharton/timber

Logging example:
https://github.com/piwik/piwik-sdk-android/blob/c081b54efb5048caa3636fe3a06c0726ff991a73/exampleapp/src/main/java/com/piwik/demo/DemoApp.java#L39

PattaFeuFeu:
@d4rken By switching from `compile` to `implementation` for Timber—as opposed to using `api`—, Timber no longer gets included as a library that can be used by consumers of the SDK. So not even the DemoApp would work. Are you planning on changing that or should I import Timber in addition to the SDK in order to be able to use the `Timber.plant(new Timber.DebugTree());` statement?

d4rken:
I'm not sure yet, I think a case could be made for both.

In any case it doesn't hurt if you manually include it for the time being.

koteswaradk:
if any one implemented PIWIk for android app please help here. i am not able to implement as there is no any errors but not showing  tracking data in y portal

PR #207, opened 2018-04-26
Fix transitive dependency on timber leading to ClassNotFoundException.

d4rken:
Was an issue with the bintray upload script.


d4rken:
Closes #204  and #184

PR #215, opened 2018-07-07
Update README.md

huangzy-android:
Dear matomo team,

I got an error when I use the SDK in my kotlin-java mixed project which jvmTarget is 1.7,
and cost lots of my time to resolve it ,  I think this change can help some one like me.


d4rken:
https://github.com/matomo-org/piwik-sdk-android/wiki/Troubleshooting

Issue #218, opened 2018-09-06
Campaign tracking


ThomasGoehringer:
I tried to add campaign tracking to my Android app. As there is no explicit method for that I tried to pass my own TrackMe object on the track method. On the object I set QueryParams.CAMPAIGN_NAME and QueryParams.CAMPAIGN_KEYWORD. However I can't find these values being tracked as campaign parameters.
Am I doing something wrong or is this not supported yet?

d4rken:
Have you checked whether the parameters are transmitted?

It's also possible that just those two keywords are not enough for the server, e.g. you'd need to also send in a tracked event or page view.

ThomasGoehringer:
I think it has to be set like the user ID. I'll have a try extending the SDK and create a PR if it works.

d4rken:
> _rcn — The Campaign name (see Tracking Campaigns). Used to populate the Referrers > Campaigns report. Note: this parameter will only be used for the first pageview of a visit.
> _rck — The Campaign Keyword (see Tracking Campaigns). Used to populate the Referrers > Campaigns report (clicking on a campaign loads all keywords for this campaign). Note: this parameter will only be used for the first pageview of a visit.

You could be right about that. This reads like you need to set these parameters on the initial trackme that's the first contact with the API for a visit.

ThomasGoehringer:
Have a look at my PR. This works for me.

d4rken:
@ThomasGoehringer I think we can close this, right?

PR #219, opened 2018-09-11
Campaign tracking

ThomasGoehringer:
#218 
I added a method for setting a campaign. This is then sent to Matomo as base params like the user ID.

### DIFF
--- a/piwik-sdk/src/main/java/org/piwik/sdk/Tracker.java
+++ b/piwik-sdk/src/main/java/org/piwik/sdk/Tracker.java
@@ -343,6 +343,22 @@ public class Tracker {
         return mDefaultTrackMe.get(QueryParams.VISITOR_ID);
     }
 
+    /**
+     * Sets the marketing campaign for this visit if a user opens your app for example because of an
+     * ad or a newsletter. This is used to populate the <i>Referrers > Campaigns</i> report.
+     * @param campaignName the name of the campaign
+     * @param keyword      the keyword of the campaign
+     */
+    public Tracker setCampaign(String campaignName, String keyword) {
+        mDefaultTrackMe.set(QueryParams.CAMPAIGN_NAME, campaignName);
+        mDefaultTrackMe.set(QueryParams.CAMPAIGN_KEYWORD, keyword);
+        return this;
+    }
+
+    public String getCampaign() {
+        return mDefaultTrackMe.get(QueryParams.CAMPAIGN_NAME) + " - " + mDefaultTrackMe.get(QueryParams.CAMPAIGN_KEYWORD);
+    }
+
     private static final Pattern PATTERN_VISITOR_ID = Pattern.compile("^[0-9a-f]{16}$");
 
     private boolean confirmVisitorIdFormat(String visitorId) throws IllegalArgumentException {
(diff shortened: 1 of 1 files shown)


d4rken:
I'd like to keep the `Tracker` as non-specific to certain use-cases as possible, meaning as few "set/get XY" methods as possible, hence the whole abstraction through `TrackMe`'s.

Ideally we would add a few helper methods here: 
https://github.com/matomo-org/piwik-sdk-android/blob/dev/piwik-sdk/src/main/java/org/piwik/sdk/extra/TrackHelper.java#L114
then it is clear that these two parameters need to be used with a tracked screen view.

In your current solution we also send the campaign fields on every request, but afaik we only need to send the parameter once, with the first tracked screen.

What do you think?

ThomasGoehringer:
I would keep it as is for two reasons:

1. Campaign tracking is most commonly used when your app is opened via a deeplink. The deeplink can then open the home screen, a product page etc. What I want to say is that you don't know which screen is opened first and so you don't know where to add a `.campaign()` like `.title()` or `.dimension()`.
2. Under https://developer.matomo.org/api-reference/tracking-api the parameters for campaigns are under **Optional User info** just like the user ID for example.

This is just my oppinion and if you know how to tackle my first point I'm open for suggestions :)

d4rken:
> Campaign tracking is most commonly used when your app is opened via a deeplink.

Is your use-case related to deep linking? AFAIK you usually catch deep-linking directly via an `Activity`. Where in your code would you call `Tracker.setCampaign...`?

>  The deeplink can then open the home screen, a product page etc. What I want to say is that you don't know which screen is opened first and so you don't know where to add a .campaign() like .title() or .dimension().

True, but don't you need to track a screen view in any of the activities anyways? If there is a common pattern to this, we could add a method to `TrackHelper` that directly takes an intent and checks it for deep-linking, otherwise an extension of `TrackHelper` for your own usecase sound like the cleanest solution to me.  Something like:

```
TrackHelper.track()
    .screen(....)
    .tryTrackDeepLink(intent)
    .with(getTracker());
```
> Under https://developer.matomo.org/api-reference/tracking-api the parameters for campaigns are under Optional User info just like the user ID for example.

I don't understand the point, can you elaborate? My issue is that the parameter only needs to be sent once, sending it everytime is wasted traffic.

ThomasGoehringer:
Sorry for the late answer.

> Is your use-case related to deep linking? AFAIK you usually catch deep-linking directly via an Activity. Where in your code would you call Tracker.setCampaign...?

We're following a one Activity, multiple Fragments approach and so we always catch deeplinks in our MainActivity and decide on where they should lead in a separate helper class. That's where I'm calling `Tracker.setCampaign()`.

I agree that sending it everytime creates unnecessary traffic but as I described you don't always track a screen view in the Activity opened by a deeplink and so I don't know if your approach is possible for everyone.

ThomasGoehringer:
> The most clean solution to me would be to give your navigation class a method like `setDeepLinkOrigin` or so, which is then consumed inside the navigator class or whereever you are calling the actual `TrackMe.track`.

I like this idea. Then I could go with a solution where the campaign is tracked with a screen like
```Java
TrackHelper.track()
    .screen(...)
    .campaign("CampaignName", "CampaignKeyword")
    .with(getTracker());
```
This could later be extended to work with Intents, too. If this is a good compromise for you I'll make a new PR.

d4rken:
> This could later be extended to work with Intents, too. If this is a good compromise for you I'll make a new PR.

Sounds good. Please also add a test for this.

Issue #255, opened 2019-03-26
Wrong usage of user id, vistor id is not persisted


eyeled:
The documentation at https://matomo.org/docs/user-id/ states that "The User ID must be the same for a given user across all her devices and browsers." Generating a random user id inside the SDK contradicts this. I think the user id should not be generated, but it should be left to the app to set it if applicable.
On the other hand the visitor is regenarated every time a Tracker instance is created. Quoting from the documentation again:  "Every unique visitor must be assigned a different ID and this ID must not change after it is assigned."

d4rken:
> The documentation at https://matomo.org/docs/user-id/ states that "The User ID must be the same for a given user across all her devices and browsers." Generating a random user id inside the SDK contradicts this. I think the user id should not be generated, but it should be left to the app to set it if applicable.

The user id is generated once then persisted per app install, the default is random because it's good for privacy and allows the sdk to support tracking out of the box. You can easily set the user-id to an email if that's what you want. If we don't do this, then every time the app is opened, it count's as a new user towards your stats. Usually devices with apps are used by a single user.

> On the other hand the visitor is regenarated every time a Tracker instance is created. Quoting from the documentation again: "Every unique visitor must be assigned a different ID and this ID must not change after it is assigned."

The visitor id ties all tracked events toegether to make up a "session", which usually lasts for the duration the app is opened, so persisting it per tracker instance (e.g. while the app is in memory) seems to be a good choice to me.


I don't see how this is being used wrong?

eyeled:
In the first place I stumbled upon this because the iOS SDK behaves differently than the Android SDK.
In the iOS SDK by default a visitor ID is randomly generated and persisted. No user ID is set by default.
In order to recognize returning visitors the visitor ID needs to stay the same. Cf. https://matomo.org/faq/general/faq_21418/
Regenerating it when the app is started again makes no sense to me. You won't recognize returning visitors.
The documentation states that the user ID must be the same for a given user *across all her devices and browsers*. That will never be the case for a generated random UUID.
So IMHO the iOS approach is how it should be done.

d4rken:
> Regenerating it when the app is started again makes no sense to me. You won't recognize returning visitors.

AFAIK the server recognizes returning visitors also by their user-id too. At least it seems to work in my console. Are you sure that returning visitors are not tracked correctly with the Android SDK?

Your explanation makes sense though, looking at the commit history it seems like it just was always the way it currently is since 2014.

[...]

PR #256, opened 2019-04-04
Align user-id / visitor-id behavior with other SDKs.

d4rken:
Generate and persist a default visitor-id.
Don't generate a default user-id by.

See #255


d4rken:
* Existing installs keep the same user-id.
* New installs will no longer get a user-id.
* All installs start using a persisted visitor-id.

PR #262, opened 2019-07-23
Specify cookies for authenticated communication with Matomo APIs

krzysztofolszewski86:
We have a case, when we set up our own Matomo API and it has to be behind a corporate proxy. Our mobile app performs authentication, but Matomo communication did not use generated cookies. As a result, no request to the API was possible and no statistics were sent. I've added a way to provide cookies for use in that communication.
While this is a niche case, the modification is small and it may be of use to somebody else.

### DIFF
--- a/exampleapp/src/main/java/org/matomo/demo/DemoApp.java
+++ b/exampleapp/src/main/java/org/matomo/demo/DemoApp.java
@@ -48,6 +48,9 @@ public class DemoApp extends MatomoApplication {
         // String userEmail = ....preferences....getString
         // getTracker().setUserId(userEmail);
 
+        // Set cookies, if your Matomo API is locked behind authorization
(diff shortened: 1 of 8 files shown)


d4rken:
Thanks for the PR @krzysztofolszewski86 looks good, but it seems your changes broke a few unit-tests.

What do you think about altering your proposed API to not be Cookie specific but instead allow for generic header additions. The Cookie related calls could be moved into an extra CookieHelper class that uses this new API.

krzysztofolszewski86:
Hi Matthias

Sorry for not responding sooner. I've been on a vacation.

I like your suggestion. I'll try to find some time to alter my code into
what you suggested, and to fix that broken unit test :).

pt., 26 lip 2019 o 15:15 Matthias Urhahn <notifications@github.com>
napisał(a):

PR #274, opened 2019-11-16
Simplify build

hannesa2:


d4rken:
Since when can this be omitted, and which version will it use?

hannesa2:
The most recent installed (on CI)

PR #279, opened 2019-12-09
Only load preferences when needed and do not use them on MainThread

Tolriq:
Avoid startup slowdown / ANR.

Despite using apply, edit and put require the preferences to be fully loaded and as disk based can be slow or ANR when used on main thread.

### DIFF
--- a/tracker/src/main/java/org/matomo/sdk/extra/InstallReferrerReceiver.java
+++ b/tracker/src/main/java/org/matomo/sdk/extra/InstallReferrerReceiver.java
@@ -3,7 +3,6 @@ package org.matomo.sdk.extra;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
-import android.content.SharedPreferences;
 
 import org.matomo.sdk.Matomo;
 
@@ -34,12 +33,15 @@ public class InstallReferrerReceiver extends BroadcastReceiver {
             Timber.tag(TAG).d("Dropping forwarded intent");
             return;
         }
-        SharedPreferences preferences = Matomo.getInstance(context.getApplicationContext()).getPreferences();
         if (intent.getAction().equals(REFERRER_SOURCE_GPLAY)) {
             String referrer = intent.getStringExtra(ARG_KEY_GPLAY_REFERRER);
             if (referrer != null) {
-                preferences.edit().putString(PREF_KEY_INSTALL_REFERRER_EXTRAS, referrer).apply();
-                Timber.tag(TAG).d("Stored Google Play referrer extras: %s", referrer);
+                final PendingResult result = goAsync();
+                new Thread(() -> {
+                    Matomo.getInstance(context.getApplicationContext()).getPreferences().edit().putString(PREF_KEY_INSTALL_REFERRER_EXTRAS, referrer).apply();
+                    Timber.tag(TAG).d("Stored Google Play referrer extras: %s", referrer);
+                    result.finish();
+                }).start();
             }
         }
         // Forward to other possible recipients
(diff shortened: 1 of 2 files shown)


tracker/src/main/java/org/matomo/sdk/extra/InstallReferrerReceiver.java
```diff
@@ -34,12 +33,13 @@ public void onReceive(Context context, Intent intent) {
             Timber.tag(TAG).d("Dropping forwarded intent");
             return;
         }
-        SharedPreferences preferences = Matomo.getInstance(context.getApplicationContext()).getPreferences();
```
d4rken: Moving the pref init into the IF scope is good :+1:

tracker/src/main/java/org/matomo/sdk/extra/InstallReferrerReceiver.java
```diff
@@ -34,12 +33,13 @@ public void onReceive(Context context, Intent intent) {
...
         if (intent.getAction().equals(REFERRER_SOURCE_GPLAY)) {
             String referrer = intent.getStringExtra(ARG_KEY_GPLAY_REFERRER);
             if (referrer != null) {
-                preferences.edit().putString(PREF_KEY_INSTALL_REFERRER_EXTRAS, referrer).apply();
-                Timber.tag(TAG).d("Stored Google Play referrer extras: %s", referrer);
+                new Thread(() -> {
```
d4rken: Launching a thread is costly too.
The `apply` is already async, it changes an internal map, and IO happens offthread already. The cost for updating a map is negligible.

So the question becomes whether it is necessary that SharedPreferences are initialized on an extra thread. I have never found that be necessary. Is this something you have experienced?

tracker/src/main/java/org/matomo/sdk/extra/InstallReferrerReceiver.java
```diff
@@ -34,12 +33,13 @@ public void onReceive(Context context, Intent intent) {
...
+                    Matomo.getInstance(context.getApplicationContext()).getPreferences().edit().putString(PREF_KEY_INSTALL_REFERRER_EXTRAS, referrer).apply();
+                    Timber.tag(TAG).d("Stored Google Play referrer extras: %s", referrer);
+                }).start();
```
d4rken: **If** we move this action off the broadcast receivers thread, we should use `goAsync()`, because we might get undefined behavior, if the broadcastreceiver finishes execution before our thread is finished.

tracker/src/main/java/org/matomo/sdk/extra/InstallReferrerReceiver.java
Tolriq: Yes I've seen quite a few times it and use this fix in prod since a few months.

The problem is not the map but how SharedPref works, when you initialize them a new thread is started to load the xml and every access is blocked via a lock. So until the preferences are loaded the edit and the putString will lock waiting for the load to be finished.
Load that can take any amount of time as disk based and OS queue can be full.

Starting a thread is way less costly than an main thread block :)

Anyway in all cases you might also want to read : https://android-developers.googleblog.com/2019/11/still-using-installbroadcast-switch-to.html as in a few month this will not work anymore at all and the other way will be mandatory.

d4rken:
Want to do a PR for the new API? 😁

Tolriq <notifications@github.com> schrieb am Mo., 9. Dez. 2019, 14:26:

Tolriq:
For the moment I lack time :( I did a quick pass on my customs Aars and pushed my fixes here since you have finally migrated to AndroidX :)

Anyway for the moment this still works and still need a fix if I want to remove my custom aars :)
I can add the goAsync if you agree about the issue. (But never saw an issue as the context used is the application context so never saw the thread unable to start before receiver ending)

d4rken:
I have not experienced such an ANR, but I trust your judgement. We can add
the thread.

I'd like the async to be added. Afaik it's not the context that's the
issue, but that the system may kill the app process if it only was launched
for the receiver, so the app process may be killed before our thread has
finished (which we now agree on, can block for theoretically arbitrary
times).

Tolriq <notifications@github.com> schrieb am Mo., 9. Dez. 2019, 14:38:

Tolriq:
PR updated.

I've added a quick hack to fix the tests as could not find an quick and easy official way to wait on the pending result.

PR #286, opened 2020-02-07
Allow extending Matomo class

veehaitch:
Currently, the constructor of the `Matomo` class is `private` disallowing the extension and hence the modification of the class. This pull request changes the accessibility to `protected`.

### DIFF
--- a/tracker/src/main/java/org/matomo/sdk/Matomo.java
+++ b/tracker/src/main/java/org/matomo/sdk/Matomo.java
@@ -46,7 +46,7 @@ public class Matomo {
         return sInstance;
     }
 
-    private Matomo(Context context) {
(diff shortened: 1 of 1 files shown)


d4rken:
Why do you need to extend it?

veehaitch:
Hey @d4rken, thanks for reaching out! In our case, we want to change the access to the Shared Preferences.

d4rken:
I don't think allowing extension here is a good choice for this library. It would make future changes difficult without causing breaking API changes for people who extended this class. Allowing extension on a class that is designed as singleton is also not without pitfalls.

For now you could fork the repo and make this change locally if needed.
The clean solution would be to offer a type of storage provider interface, feel free to create an issue ticket for that. Please include your specific requirements.

PR #293, opened 2020-08-10
Github actions

hannesa2:
advantage:
* faster
* easy to run Espresso tests, but in this repo there are no Espresso test, so I didn't run it


hannesa2:
If you like it, you could remove Travis

d4rken:
I like it, but I don't want to change it "just because", it would need to provide a specific benefit that we are currently missing.

Currently the CodeCoverage is broken again, would this fix it? Currently don't have time to investigate Travis again :grin:...

hannesa2:
I only applied the same commands like in Travis. Broken or not, at least there is no error.

I at least move complete from Travis to Github actions because of the two reasons https://github.com/matomo-org/matomo-sdk-android/pull/293#issue-465407473

d4rken:
Will fix this up on a different branch.

PR #297, opened 2020-10-30
update DATETIME_OF_REQUEST to unix timestamp

Isaacwhyuenac:
Would want to have the DATETIME_OF_REQUEST(cdt) query params in unix timestamp format because it includes milliseconds which help ordering of events much better. Also, it aligns with other version of sdk, like the iOS and web versions.


hannesa2:
Why did you closed it ? For me it makes sense

Isaacwhyuenac:
Because I find a similar issue https://github.com/matomo-org/matomo-sdk-android/issues/54 that is discussing similar things but never closed. Maybe reopen the pull request

Issue #309, opened 2021-07-10
Device information only report for first session per process


kevinslashslash:
Device information, such as resolution, user agent and language, only get reported for the first session causing "unknown" screen resolution and "Unknown" device brands and "Generic Mobile device" device models.
Steps to reproduce:
Setup backend and Demo App
In Demo App press Track Screen View then Dispatch Now
Verify this is reporting in Visits Log or Real-time. See that the device info is set (specifically resolution as brand sometimes makes it through).
Now without having exited the Demo App, press Track Screen View and then Dispatch Now again.
This time check the reporting and the resolution will say "unknown"

This is likely related to this code:
https://github.com/matomo-org/matomo-sdk-android/blob/3080d186902968bde02a197274fe9d9efd91bb47/tracker/src/main/java/org/matomo/sdk/Tracker.java#L441

`mLastEvent` being non-null is being used as a hint that this is part of an existing session, but that isn't necessarily true. In the case of the demo app, Track Screen View is specifically starting a new session, but it could also happen by timeout.
I think it'd be most appropriate to move setting of three parameters (`SCREEN_RESOLUTION`, `USER_AGENT`, `LANGUAGE`) into their own method and call it from `track` if `SESSION_START` is set.
Alternatively if the way "Track Screen View" is starting a new session just by setting `SESSION_START` is inappropriate, that code should be changed and the three parameters could be part of `injectInitialParams` instead.

d4rken:
What do you think about extending the newSession check, to also check for the new session flag in the `TrackMe`:

https://github.com/matomo-org/matomo-sdk-android/blob/51181f41060d534fe59595455518266cd2a441d5/tracker/src/main/java/org/matomo/sdk/Tracker.java#L451-L456

> I think it'd be most appropriate to move setting of three parameters (SCREEN_RESOLUTION, USER_AGENT, LANGUAGE) into their own method and call it from track if SESSION_START is set.

Maybe we simplify this and also call it for every  inject base parameters call, starting to seem like this optimization is more trouble than it's worth... :thinking:

kevinslashslash:
The fix in my pull request does fix the issue as described, however it doesn't eliminate sessions with no device information.

I think another way this can happen is that the session start event gets purged from the cache but the rest of the session doesn't. This could be addressed either by injecting the parameters to the session at dispatch time, which seems wrong as it's mixing responsibilities, injecting on every event, or altering dispatch logic to not even submit sessions when the session start was dropped.

[...]

PR #310, opened 2021-07-12
Fix Device information only report for first session per process

kevinslashslash:
As discussed in #309 this addresses the issue of the fingerprint parameters not being added for new sessions, causing "unknown" and "Generic" device information

Also there are some additional commits for minor cleanup of Tracker.java


d4rken:
Continued in #319

PR #316, opened 2021-07-19
Dependabot to keep repo up to date

hannesa2:


d4rken:
I don't think we need to spam this repo with dependabot alerts 😀

hannesa2:
Yes, let's stay on outdated stuff and don't move 😀

d4rken:
If we need something for a fix, ofc we can update.

But every time a bot says there is a new update? 🤔 
I'm not getting paid to maintain this and want to keep overhead to a minimum 🍻 .

If you want to bump a version for good reason, feel free to do so ❤️ .

PR #319, opened 2021-08-02
Fix server side device fingerprinting failing in edge cases.

d4rken:
Continuation of #310 
Inject device infos for fingerprinting into every packet.


d4rken:
@kevinslashslash what do you think?

kevinslashslash:
LGTM

PR #328, opened 2022-07-08
Update # FAQ Links

Starker3:
A recent knowledge base update has changed how FAQ links work on matomo.org

Links with `#` in the link no longer work for guides or FAQs


d4rken (approved):
Thanks!

PR #332, opened 2022-07-10
Fix tests and simplify

hannesa2:
close #330 

Now `./gradlew test` works locally
image


tracker/src/test/java/org/matomo/sdk/extra/CustomVariablesTest.java
```diff
@@ -1,6 +1,5 @@
 package org.matomo.sdk.extra;
 
-import org.apache.maven.artifact.ant.shaded.StringUtils;
```
hannesa2, line 3: I'm happy to remove this

d4rken: Me2, weird to have that :grin:

d4rken (approved):
Nice!

PR #334, opened 2022-07-10
Release job and automatic changelog

hannesa2:
A new release needs now only a new tag, eg
```
git tag -a 1.0 -m "text"
git push --tags
```

Then the release pipeline starts and generates a changelog, depended on labels of pull request.
This means a label on pull request makes sense now, if you want to see it in release changelog,

All is untested and copy&paste from https://github.com/MikeOrtiz/TouchImageView where it works like this

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,47 @@
+name: Changelog vs Release
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Build Changelog
+        id: github_release
+        uses: mikepenz/release-changelog-builder-action@main
+        with:
+          configuration: ".github/changelog-configuration.json"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install JDK ${{ matrix.java_version }}
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'adopt'
+          java-version: 11
+      - name: Build project
+        run: ./gradlew assembleRelease
+        env:
+          VERSION: ${{ github.ref }}
+      - name: Get the version
+        id: tagger
+        uses: jimschubert/query-tag-action@v2
+        with:
+          skip-unshallow: 'true'
+          abbrev: false
+          commit-ish: HEAD
+      - name: Create Release
+        uses: softprops/action-gh-release@v1
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+          configuration: ".github/changelog-configuration.json"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install JDK ${{ matrix.java_version }}
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'adopt'
+          java-version: 11
+      - name: Install Android SDK
+        uses: malinskiy/action-android/install-sdk@release/0.1.2
```
d4rken, line 29: Is this needed?
The Github `ubuntu-latest` machine already has the Android SDK installed.

e.g. the other workflow runs without it:
https://github.com/matomo-org/matomo-sdk-android/blob/7216a3227cd225abd2850ba9615f0d05194079ca/.github/workflows/pull-request-ci.yml#L14-L26

d4rken (commented):
Neat, looking forward to see the changelog generator in action :sunglasses:.

One small question, see comment.

Findus23:
Something like https://github.com/matomo-org/matomo/pull/17809 might also be useful here

.github/workflows/release.yml
hannesa2: Ok, I give it a try

hannesa2:
Because I want to test it, I introduce a pre-release ability

.github/workflows/release.yml
hannesa2: Indeed it's not needed

hannesa2:
> Because I want to test it, I introduce a pre-release ability

Pre-relase is working, but changelog not

.github/workflows/release.yml
d4rken: Nice :slightly_smiling_face:

hannesa2:
I made a test release https://github.com/hannesa2/matomo-sdk-android/releases in my fork, and there at least, it was working

d4rken (approved):
Does it only work for master maybe?

hannesa2:
> Does it only work for master maybe?

No, it works on a branch as well. Even pull request with merge commit and  pull request with rebased commit.
image

Here are my tests https://github.com/hannesa2/matomo-sdk-android/releases

I've no clue, why the tests in this repository where without changelog. Maybe merge it and lighten a candle that next time  on real release it will work on this repo as well ?

hannesa2:
I removed the pre-release because it can't show tag in caption 
image

Now it works like this gain. But without the pre-release ability
image

Please see my tests https://github.com/hannesa2/matomo-sdk-android/releases

PR #337, opened 2022-07-11
Remove lint warnings

hannesa2:
I followed the suggestion from lint

### DIFF
--- a/tracker/src/main/java/org/matomo/sdk/Matomo.java
+++ b/tracker/src/main/java/org/matomo/sdk/Matomo.java
@@ -30,7 +30,7 @@ public class Matomo {
     private static final String TAG = Matomo.tag(Matomo.class);
     private static final String BASE_PREFERENCE_FILE = "org.matomo.sdk";
 
-    @SuppressLint("StaticFieldLeak") private static Matomo sInstance;
+    @SuppressLint("StaticFieldLeak") private static volatile Matomo sInstance;
 
     private final Map<Tracker, SharedPreferences> mPreferenceMap = new HashMap<>();
     private final Context mContext;
--- a/tracker/src/main/java/org/matomo/sdk/TrackerBuilder.java
+++ b/tracker/src/main/java/org/matomo/sdk/TrackerBuilder.java
@@ -7,8 +7,8 @@ import java.net.URL;
  * Configuration details for a {@link Tracker}
  */
 public class TrackerBuilder {
-    private String mApiUrl;
-    private int mSiteId;
+    private final String mApiUrl;
+    private final int mSiteId;
     private String mTrackerName;
     private String mApplicationBaseUrl;
 
--- a/tracker/src/main/java/org/matomo/sdk/extra/DownloadTracker.java
+++ b/tracker/src/main/java/org/matomo/sdk/extra/DownloadTracker.java
@@ -26,7 +26,6 @@ public class DownloadTracker {
     private final Object mTrackOnceLock = new Object();
     private final PackageManager mPackMan;
     private final SharedPreferences mPreferences;
-    private final Context mContext;
     private final boolean mInternalTracking;
     private String mVersion;
     private final PackageInfo mPkgInfo;
@@ -68,8 +67,6 @@ public class DownloadTracker {
                     mPackageInfo = null;
                 }
             }
-
-            public ApkChecksum(PackageInfo packageInfo) {mPackageInfo = packageInfo;}
 
             @Override
             public boolean isIntensiveWork() {
@@ -128,7 +125,7 @@ public class DownloadTracker {
 
     public DownloadTracker(Tracker tracker, @NonNull PackageInfo packageInfo) {
         mTracker = tracker;
-        mContext = tracker.getMatomo().getContext();
+        Context mContext = tracker.getMatomo().getContext();
         mPreferences = tracker.getPreferences();
         mPackMan = tracker.getMatomo().getContext().getPackageManager();
         mPkgInfo = packageInfo;
--- a/exampleapp/src/main/java/org/matomo/demo/DemoApp.java
+++ b/exampleapp/src/main/java/org/matomo/demo/DemoApp.java
@@ -18,7 +18,6 @@ import org.matomo.sdk.extra.TrackHelper;
 import timber.log.Timber;
 
 public class DemoApp extends MatomoApplication {
-    private DimensionQueue mDimensionQueue;
 
     @Override
     public TrackerBuilder onCreateTrackerConfig() {
@@ -55,7 +54,7 @@ public class DemoApp extends MatomoApplication {
         // i.e. "http://org.matomo.demo:1/com.android.vending"
         // getTracker().download();
 
-        mDimensionQueue = new DimensionQueue(getTracker());
+        DimensionQueue mDimensionQueue = new DimensionQueue(getTracker());
 
         // This will be send the next time something is tracked.
         mDimensionQueue.add(0, "test");
--- a/exampleapp/src/main/java/org/matomo/demo/SettingsActivity.java
+++ b/exampleapp/src/main/java/org/matomo/demo/SettingsActivity.java
@@ -30,7 +30,7 @@ public class SettingsActivity extends Activity {
 
     private void refreshUI(final Activity settingsActivity) {
         // auto track button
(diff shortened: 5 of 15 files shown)


tracker/src/main/java/org/matomo/sdk/TrackMe.java
```diff
@@ -94,37 +93,33 @@ public synchronized boolean has(@NonNull QueryParams queryParams) {
      *
      * @param key   type
      * @param value value
-     * @return this (for chaining)
      */
-    public synchronized TrackMe trySet(@NonNull QueryParams key, int value) {
-        return trySet(key, String.valueOf(value));
+    public synchronized void trySet(@NonNull QueryParams key, int value) {
```
d4rken, line 97: Changes public API, see above.

tracker/src/main/java/org/matomo/sdk/TrackMe.java
```diff
@@ -94,37 +93,33 @@ public synchronized boolean has(@NonNull QueryParams queryParams) {
...
     /**
      * Only sets the value if it doesn't exist.
      *
      * @param key   type
      * @param value value
-     * @return this (for chaining)
      */
-    public synchronized TrackMe trySet(@NonNull QueryParams key, float value) {
-        return trySet(key, String.valueOf(value));
+    public synchronized void trySet(@NonNull QueryParams key, float value) {
```
d4rken, line 107: Changes public API, see above.

tracker/src/main/java/org/matomo/sdk/TrackMe.java
```diff
@@ -94,37 +93,33 @@ public synchronized boolean has(@NonNull QueryParams queryParams) {
...
+        trySet(key, String.valueOf(value));
     }
 
-    public synchronized TrackMe trySet(@NonNull QueryParams key, long value) {
-        return trySet(key, String.valueOf(value));
+    public synchronized void trySet(@NonNull QueryParams key, long value) {
```
d4rken, line 111: Changes public API, see above.

tracker/src/main/java/org/matomo/sdk/TrackMe.java
```diff
@@ -94,37 +93,33 @@ public synchronized boolean has(@NonNull QueryParams queryParams) {
...
 
     /**
      * Only sets the value if it doesn't exist.
      *
      * @param key   type
      * @param value value
-     * @return this (for chaining)
      */
-    public synchronized TrackMe trySet(@NonNull QueryParams key, String value) {
+    public synchronized void trySet(@NonNull QueryParams key, String value) {
```
d4rken, line 121: Changes public API, see above.

tracker/src/main/java/org/matomo/sdk/TrackerBuilder.java
```diff
@@ -43,9 +43,8 @@ public int getSiteId() {
     /**
      * A unique name for this Tracker. Used to store Tracker settings independent of URL and id changes.
      */
-    public TrackerBuilder setTrackerName(String name) {
+    public void setTrackerName(String name) {
```
d4rken, line 46: Changes public API (chaining calls on a builder object `Builder().doA().doB()`

tracker/src/main/java/org/matomo/sdk/TrackerBuilder.java
```diff
@@ -58,9 +57,8 @@ public String getTrackerName() {
      *
      * @param domain your-domain.com
      */
-    public TrackerBuilder setApplicationBaseUrl(String domain) {
+    public void setApplicationBaseUrl(String domain) {
```
d4rken, line 60: Changes public API

tracker/src/main/java/org/matomo/sdk/extra/CustomVariables.java
```diff
@@ -70,9 +70,8 @@ public CustomVariables(@Nullable String json) {
         }
     }
 
-    public CustomVariables putAll(CustomVariables customVariables) {
+    public void putAll(CustomVariables customVariables) {
```
d4rken, line 73: Changes public API

tracker/src/main/java/org/matomo/sdk/extra/DownloadTracker.java
```diff
@@ -88,12 +85,6 @@ public String buildExtraIdentifier() {
             }
         }
 
-        /**
-         * Custom exta identifier. Supply your own \o/.
-         */
-        abstract class Custom implements Extra {
-        }
-
```
d4rken, line 96: I agree, this is kind of superfluos, but it's part of the public API and some devs might have extended this.

d4rken (changes requested):
Some of the changes affect the public API that devs might be using (see comment), the rest looks good, thanks for the cleanup :+1:

tracker/src/main/java/org/matomo/sdk/extra/DownloadTracker.java
hannesa2: Keeping this unused stuff is a kind of collect a lot waste in your base ground.

tracker/src/main/java/org/matomo/sdk/Matomo.java
```diff
@@ -30,7 +30,7 @@ public class Matomo {
     private static final String TAG = Matomo.tag(Matomo.class);
     private static final String BASE_PREFERENCE_FILE = "org.matomo.sdk";
 
-    @SuppressLint("StaticFieldLeak") private static Matomo sInstance;
+    @SuppressLint("StaticFieldLeak") private static volatile Matomo sInstance;
```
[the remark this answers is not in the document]
hannesa2: image

tracker/src/main/java/org/matomo/sdk/extra/DownloadTracker.java
d4rken: We can remove it, but then with a major version bump and a little note about migration in the changelog (e.g. don't extend `Custom`, just implement `Extra`).

tracker/src/main/java/org/matomo/sdk/Matomo.java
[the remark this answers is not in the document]
d4rken: Visibility issue when accessing it from multiple threads should be handled by `synchronized` already :thinking: 

Hm, it doesn't hurt.  :shrug:

d4rken (approved):
Thank you!

tracker/src/main/java/org/matomo/sdk/TrackMe.java
```diff
@@ -30,9 +30,8 @@ public TrackMe(TrackMe trackMe) {
     /**
      * Adds TrackMe to this TrackMe, overriding values if necessary.
      */
-    public TrackMe putAll(@NonNull TrackMe trackMe) {
+    public void putAll(@NonNull TrackMe trackMe) {
```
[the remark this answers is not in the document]
hannesa2: Sure, I as a customer will only use stuff I find in sample app. It's just a note.

> The linter doesn't know it's an SDK and code is used outside the project

yes, maybe annotate it with "ignore warning" is a way to go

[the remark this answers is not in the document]
hannesa2: Till now, I ignored my wish to kotlize it.
Kotlin has some wonderful mechanism for a sdk, e.g `internal`

hannesa2:
Nice, we come closer 
image

hannesa2:
But this interface way isn't done

image

d4rken:
> This keeping unused classes because someone could use it, is an afford without benefit.

Why do you keep coming back to this? :smile: 
I know it's not great, I probably just forgot about that class. I had no one to review my code.
It's out there now and we shouldn't break the users app "just because it's not clean".
Sorry for not writing the perfect SDK? :thinking: 

> But this interface way isn't done

Not everything needs to be a firebase-level SDK with shovels of abstraction.
This is volunteer work and the SDK worked fine for my use-cases, so at that time: Why make things more complicated if there is no requirement?
A rewrite to Kotlin wasn't on the horizon at that point.

Interfaces can still be introduced now if desired. As said, you can make a v5, with everything new and better if you want to  :+1:.

hannesa2:
Sorry, those were just a few thoughts on the Kotlin transformation.
You are right, house holding of our time is also important.

PR #362, opened 2022-07-29
Update example app with new demo server

d4rken:
The previous one was lost? when it went `Piwik` -> `Matomo`.

The new one is `demo2.matomo.org`, ID `81`.

@hannesa2 You should have gotten an invite with admin access to the server. Might have been flagged as spam?


hannesa2:
I followed all required steps and ended here 
image
To be honest, I don't know what's the purpose of it

hannesa2 (approved):

d4rken:
> To be honest, I don't know what's the purpose of it

The PR to get the demo app working correctly,
and the invite f you want to test something, you don't need to pay/host an instance yourself.

Issue #259, opened 2019-07-02
Reset UserId while logged out


JeyakumarThangaraj:
Team,

We dont find anyway to reset userId while logout. because of this, visitor id remain same for all the time whenever same user is trying to login. Kindly Advise

Thanks,
Jeyakumar

d4rken:
What about `Tracker.setUserId(...)`.

https://github.com/matomo-org/matomo-sdk-android/blob/006ed7dfb4541273adc43bbd315bcd955dc32379/tracker/src/main/java/org/matomo/sdk/Tracker.java#L334

JeyakumarThangaraj:
If i reset user Id, new visitor ID and a new visit is not created. Still existing visit only being used. Kindly advise. 
I want below to be implemented in android.

![image](https://user-images.githubusercontent.com/49137580/60824687-b38b1d00-a1b2-11e9-9ce6-9d052356f7b7.png)

d4rken:
`resetUserId` also resets the visitor-id in the [javascript tracker](https://github.com/matomo-org/matomo/blob/7a93a2b5ff683e7bb27339502a60c00add9fe26e/tests/javascript/index.php)

So to do the analog here you'd probably have to do something like:

```kotlin
tracker.setUserId(null)
tracker.setVisitorId(Tracker.makeRandomVisitorId())
```

Also see `SESSION_START` (i.e. [`new_visit`](https://developer.matomo.org/api-reference/tracking-api)), which you could set to prevent the server from attempting any visit stitching.

```kotlin
TrackHelper.track(TrackMe().set(QueryParams.SESSION_START,1))
    .screen("/").title("Main screen")
    .with(getTracker())
```

Nithiyasabariswaran:
Hi All,

All my screens are loading in a single visitor id (Android).
I have used tracker.setVisitorId(Tracker.makeRandomVisitorId()) also.

kinldy advice,
Thanks in advance.

![image](https://user-images.githubusercontent.com/12802349/60833760-afb5c580-a1c7-11e9-86e4-44ac102cbae3.png)

mattab:
> resetUserId also resets the visitor-id in the javascript tracker

Fyi @d4rken we recently changed this and no longer will resetUserId change the Visitor ID. This was done in https://github.com/matomo-org/matomo/pull/14360

hannesa2:
Solved with  #373
If not, please reopen it

Issue #261, opened 2019-07-19
Feat: Addition of a Tracker.resetTracking() call?


d4rken:
Also see #259 

"Old user logs out, new user logs in, track as new visit" seems to be a common yet problematic use-case that could be made easier by providing a call like:

```java
class Tracker {
   /* .... */ 
   void resetVisit() {
        // Null user id
        // Set new random visitor id
        // set NEW_VISIT = 1
    }
}
```

Apparently the JavaScript SDK has this, I'm not certain whether the iOS SDK has this?

`resetVisit` might be a bit missleading and something like `resetTracking` would be more appropriate in conveying the effect.

mattab:
Fyi here is how we now recommend to do this in JavaScript: https://developer.matomo.org/guides/tracking-javascript-guide#when-user-logs-out-reset-user-id And similarly documented in https://matomo.org/faq/how-to/faq_187/

it's a bit convoluted/complex but that's the best we got :)

VikasRao:
> https://matomo.org/faq/how-to/faq_187/

but how to push new_visit=1 from android, should we use Tracker.setUserId(1) or setVisitorId  or any other method?

d4rken:
```java
        TrackMe trackMe = TrackHelper.track()
                .event("exampleCategory","exampleAction")
                .build();
        trackMe.set(QueryParams.SESSION_START, 1);
        mTracker.track(trackMe);
```

hannesa2:
Solved with  #373
If not, please reopen it

PR #373, opened 2022-09-16
Add ability to reset tracker state

FabioTacke:
This adds the feature to reset the whole tracker state in order to decouple tracking sessions manually.
This way the data bound to a device right now can be reset and treat the user as a new user.


hannesa2 (approved):
Nice, even with a test for it !
If @d4rken doesn't find anything against it, I will merge it tomorrow

d4rken:
> If @d4rken doesn't find anything against it, I will merge it tomorrow

I'm okay with it if you are :+1: 

> Nice, even with a test for it ! I

:heart:

Would this PR relate to or even close #259 or #261?

PR #438, opened 2023-09-24
Hard strict mode

hannesa2:


hannesa2:
Hmm, same strict mode finds a lot locally but not on CI

Issue #468, opened 2024-01-11
Single event doesn't appear on Dashboard


nostromoo:
Hello,

We are using the android Matomo SDK version 4.1.4. We have notice that when we log one single event on a dispatchInterval period, it doesn't appear in the dashboard, but when there are several, it works fine.

After analysing the code of the Matomo  sdk, I notice that to report a single event, the SDK is doing a http GET request and if there are more than one a http POST request (see send() method in DefaultPacketSender). 

I tried to fork the Matomo SDK and modify the code to ensure that we use the POST call all the time, and it works fine.

Could you fix the issue please? Or tell us if there is a problem in our integration?

Thanks.

hannesa2:
When you fixed it, it would be fine, when you make a pull request for it

nostromoo:
I have an error when I try to push my branch : 

remote: Permission to matomo-org/matomo-sdk-android.git denied to nostromoo.
fatal: unable to access 'https://github.com/matomo-org/matomo-sdk-android.git/': The requested URL returned error: 403

I tried with username/password and accessToken login

hannesa2:
You are not allowed to push to this repository, you need first an own fork.
![image](https://github.com/matomo-org/matomo-sdk-android/assets/3314607/b9acfae0-ae61-4e3c-a8f8-4ec8906e7025)
Add this as a new remote to your repository and push it there.
With this you can make a pull request.

Here you have the long and complete story https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests

nostromoo:
Thank you for your reply, it's done :
https://github.com/matomo-org/matomo-sdk-android/pull/470

PR #470, opened 2024-01-23
Send packet only with POST method.

nostromoo:


tracker/src/main/java/org/matomo/sdk/dispatcher/PacketFactory.java
```diff
@@ -37,19 +37,11 @@ public PacketFactory(final String apiUrl) {
...
-            if (p == null) return Collections.emptyList();
-            else return Collections.singletonList(p);
-        }
-
         int packets = (int) Math.ceil(events.size() * 1.0 / PAGE_SIZE);
         List<Packet> freshPackets = new ArrayList<>(packets);
         for (int i = 0; i < events.size(); i += PAGE_SIZE) {
             List<Event> batch = events.subList(i, Math.min(i + PAGE_SIZE, events.size()));
-            final Packet packet;
-            if (batch.size() == 1) packet = buildPacketForGet(batch.get(0));
```
hannesa2, line 51: @d4rken 
When I understand right, you did in 1f8ccd6253b3e9f29cc84d4c45587931a62a10c8 this decision if it's a `get` or a `post` by asking `batch.size()`, am I right ?

Do you agree with this changes ?

d4rken: I don't remember why I did it this way :thinking:.

Might have been an optimization, i.e. `GET` having less overhead than `POST` for single events. In hindsight I don't think the overhead would be huge enough to warrant both GET and POST.

I'm neutral on this change.

@nostromoo Please consider adding some descriptions to your PRs. What made you change this?

hannesa2 (approved):
Like @d4rken I'm neutral on this change too.
We can give it a try

hannesa2:
@nostromoo 
Please test it first 
`implementation 'com.github.nostromoo:matomo-sdk-android:remove_get_packet-SNAPSHOT'`
if it comes now with desired behavior. If yes, I'll merge it.

Please keep me informed

d4rken:
Now I see the linked #468.

Did something change on the server-side? According to docs (the last time I checked) the server should accept both GET and POST. :shrug: 

This could warrant further debugging to find out why `GET` is not working. Broken in the SDK? Or an issue with the server config?

If it's the later, changing the SDK seems like the wrong way to fix it :grin:.

hannesa2:
> Now I see the linked #468.

I did it

hannesa2:
> If it's the later, changing the SDK seems like the wrong way to fix it 😁.

Any suggestion what's the right way

d4rken:
> > If it's the later, changing the SDK seems like the wrong way to fix it 😁.
> 
> Any suggestion what's the right way

Check if single `GET` requests are reaching the server, and if they are, check why the server is not processing them. Removing the `GET` and only using `POST` is fine, but shouldn't be done as a "guess" without understanding why `GET` is not working.

@mattab Did the server's behavior for `GET` change?

PR #476, opened 2024-02-07
Avoid clearing dispatcher when user changes optOut setting

### LINKED ISSUE #474: setOptOut clears dispatecher each time the method is invoked
Hi,
each time `setOptOut` method is invoked the dispatcher queue is cleared.

```java
    /**
     * Use this to disable this Tracker, e.g. if the user opted out of tracking.
     * The Tracker will persist the choice and remain disable on next instance creation.<p>
     *
     * @param optOut true to disable reporting
     */
    public void setOptOut(boolean optOut) {
        mOptOut = optOut;
        getPreferences().edit().putBoolean(PREF_KEY_TRACKER_OPTOUT, optOut).apply();
        mDispatcher.clear();
    }
```
    
When app has switch that enable / disable event tracking we can loose events history. Example scenario:

1. App collected many events during last 1 minute.
2. Uses goes to settings. 
3. User disables tracking. `setOptOut` is invoked. Dispatcher is cleared. 
4. User enables tracking. `setOptOut` is invoked. Dispatcher is cleared. 
5. User disables tracking. `setOptOut` is invoked. Dispatcher is cleared. 
6. User enables tracking. `setOptOut` is invoked. Dispatcher is cleared. 
7. No events to dispatch. 

In that case user still allows tracking but we lost all collected events. 

I think it should work this way:

1. App collected many events during last 1 minute.
2. Uses goes to settings. 
3. User disables tracking. `setOptOut` is invoked. Events collecting is disabled. Next event will not be collected.
4. User enables tracking. `setOptOut` is invoked.  Events collecting is enabled. Next event will be collected.
5. User disables tracking. `setOptOut` is invoked.  Events collecting is disabled. Next event will not be collected.
6. User enables tracking. `setOptOut` is invoked.  Events collecting is enabled. Next event will be collected.
7. All collected events when user allowed to be tracked are dispatched. 

I think this is the way iOS SDK works at the moment.

lorenc-tomasz:
PR fixes #474 

Basically setting setting setOptOut always clears dispatcher. It should set only flag and events should be omitted, not cleared.

### DIFF
--- a/tracker/src/main/java/org/matomo/sdk/Tracker.java
+++ b/tracker/src/main/java/org/matomo/sdk/Tracker.java
@@ -164,7 +164,6 @@ public class Tracker {
     public void setOptOut(boolean optOut) {
         mOptOut = optOut;
         getPreferences().edit().putBoolean(PREF_KEY_TRACKER_OPTOUT, optOut).apply();
-        mDispatcher.clear();
     }
 
     /**
--- a/tracker/src/test/java/org/matomo/sdk/TrackerTest.java
+++ b/tracker/src/test/java/org/matomo/sdk/TrackerTest.java
@@ -234,7 +234,6 @@ public class TrackerTest {
     public void testOptOut_set() {
         Tracker tracker = new Tracker(mMatomo, mTrackerBuilder);
         tracker.setOptOut(true);
-        verify(mDispatcher).clear();
         assertTrue(tracker.isOptOut());
         tracker.setOptOut(false);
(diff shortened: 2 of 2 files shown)


hannesa2:
The tests are good, but they need an adaption too

lorenc-tomasz:
Fixed. Sorry for that.

hannesa2:
https://github.com/matomo-org/matomo-sdk-android/releases/tag/4.2

PR #505, opened 2024-06-06
Show OS in CI

hannesa2:
image


hannesa2:
It helps for build time measurement 
image

PR #523, opened 2024-07-11
Force tracker timeout

hannesa2:


hannesa2:
@ari-github I don't see a timeout. How to reproduce it ?

image

PR #637, opened 2026-04-14
Update pull-request-ci.yml

AltamashShaikh:


.github/workflows/pull-request-ci.yml
```diff
@@ -1,4 +1,4 @@
-name: PullRequest
+name: PullRequest Edit
```
hannesa2, line 1: Which problem does it solve '?

sgiehl: That had been a test only.

